PR #453, opened 2022-07-09
add ClearPlayerTasks and fix ClearAnimations

AmyrAhmady:
DO NOT MERGE THIS YET
ONLY MERGE WHEN WE HAVE OUR OWN INCLUDE PACKS

This just a PR to bring some changes back from here: https://github.com/openmultiplayer/open.mp/pull/447
And in addition to that, there will be also a new native called `ClearPlayerTasks` doing what `ClearAnimations` was doing originally


Zorono:
ummm...
> ONLY MERGE WHEN WE HAVE OUR OWN INCLUDE PACKS

is this PR ready for production use ??

AmyrAhmady:
No

PR #499, opened 2022-09-10
x64

Y-Less:
I could split this in to two PRs if people don't want to read through two bits of code.  It actually adds two things:

1. Config improvements:
- Save unknown legacy options from server.cfg, so they can still be used even if they aren't recognised.  Attempts to type them based on value.
- Add special case for reading `gamemode0` etc config options in pawn, since that needs to be converted to a string array access.
2. Building pawn with x64 support:
- The weird DB natives that return pointers won't work, but they're the only ones.
- amxfile now uses IDs instead of pointers for returned values when the pointers won't fit in.  I think this also fixes the use-after-close problems people were having, as a side-effect, but this isn't enabled on the 32-bit version by default.  It could be.  Edit: It now is, so this also fixes that unfixable bug.
- `OP_SWITCH` now uses a binary search.
- There is a caveat to this.  The 64-bit upgrade for 32-bit pawn is based on the work done upstream, which still stores the 64-bit pointers to the native implementations in the header as before.  But obviously they're now twice as wide, which means they take up twice as much space as is available.  This thus clobbers the pointers to the native's names, so once a script is loaded in 64-bit you can't look up a native by name.


AGraber:
I took the liberty to add some non-code changes that were required for keeping builds and GitHub CI working:

- Small fix on CMakeLists to avoid x86_64 builds matching the x86 part of the string resulting in -m32 being added to 64bit builds on Linux
- Changes to GitHub CI workflows to simplify them while adding x64 support to them as well

AmyrAhmady (approved):

PR #567, opened 2023-01-04
Implement (Start|Stop)RecordingPlayerData

WoutProvost:
See https://github.com/openmultiplayer/server-beta-old/issues/228 and https://github.com/openmultiplayer/server-beta-old/issues/215.

Analyzed recorded data with my [rec2csv tool](https://github.com/WoutProvost/samp-rec-to-csv) after implementation for both on foot and driver types. Seems fine.

Since the header is versioned (as was in SA-MP), we can always create a new `.rec` structure if we want to, and bump the version. Then we could even load in the data from old `.rec` files, convert it and write it out using our new format. But this is of course out of scope for this PR.


ksenonadv (changes requested):

Server/Components/Recordings/recordings_main.cpp
```diff
@@ -0,0 +1,229 @@
...
+		core->getPlayers().getPlayerConnectDispatcher().addEventHandler(this);
+		NetCode::Packet::PlayerFootSync::addEventHandler(*core, &onFootRecordingHandler);
+		NetCode::Packet::PlayerVehicleSync::addEventHandler(*core, &driverRecordingHandler);
+	}
+
+	void reset() override
+	{
+		for (IPlayer* player : core->getPlayers().entries())
+		{
+			queryExtension<IPlayerRecordingData>(player)->stop();
```
ksenonadv: Missing nullptr check here.

ksenonadv (changes requested):

SDK/include/Server/Components/Pawn/pawn.hpp
```diff
@@ -57,6 +58,7 @@ struct PawnLookup
 	IMenusComponent* menus = nullptr;
 	IObjectsComponent* objects = nullptr;
 	IPickupsComponent* pickups = nullptr;
+	IRecordingsComponent* recordings = nullptr;
```
Y-Less, line 61: I think this is an ABI break, but the whole lookup system is bad and I should change it (especially since we already have a robust flexible component lookup system).

WoutProvost: Any workaround now or is this PR still blocked because of this?

Y-Less: I'll double-check, as as pointed out on discord this shouldn't be a problem - it isn't in a public interface.

Y-Less:
I've merged this.  I originally wasn't sure because the ABI tool reported a break, but I have no clue why, and then #685 was opened with the same failure.  I'm going to make a third quick PR, just to try see the report output, but then I'll close this.

PR #595, opened 2023-01-10
Bug fixes in SDK/Include/

temajm:
Small corrections to the code :)

```Before doing the merger, the conflicts described in the TODO must be resolved.```

### CODE REVIEW

File: SDK/include/values.hpp [hunk 1]
```diff
@@ -9,7 +9,11 @@ constexpr int CLASS_POOL_SIZE = 320;
 constexpr int OBJECT_POOL_SIZE = 2000;
 constexpr int OBJECT_POOL_SIZE_037 = 1000;
 constexpr int MAX_WEAPON_SLOTS = 13;
-constexpr int MAX_VEHICLE_MODELS = 611 - 400 + 1;
+constexpr int MIN_COMPONENT_ID = 1000;
+constexpr int MAX_COMPONENT_ID = 1193;
```
Y-Less, line 13: It's not common for ranges in code to be inclusive like this.  A think a define for `1194` would be more idiomatic, and avoid some `+ 1`s (e.g. `MAX_COMPONENT_ID - MIN_COMPONENT_ID + 1`), but I'm not sure what the best name for it would be.

SDK/include/values.hpp
```diff
@@ -62,33 +66,33 @@ constexpr int MAX_GAMETEXT_STYLES = 16;
...
 {
-	ACTION = 1,
-	CROUCH = 2,
-	FIRE = 4,
-	SPRINT = 8,
-	SECONDARY_ATTACK = 16,
-	JUMP = 32,
-	LOOK_RIGHT = 64,
-	HANDBRAKE = 128,
+	ACTION = 0b1 << 0,
```
Y-Less: The `0b` is logically correct, but since everything is just `0b1` I think not using `1` is just visual clutter.

SDK/include/Server/Components/Vehicles/vehicle_colours.hpp
```diff
@@ -698,7 +698,7 @@ inline void getRandomVehicleColour(int modelid, int& colour1, int& colour2)
 	return getRandomVehicleColour(modelid, colour1, colour2, colour3, colour4);
 }
 
-inline uint32_t carColourIndexToColour(int index, uint32_t alpha = 0xFF)
+inline uint32_t carColourIndexToColour(int index, uint32_t alpha = 0xFF) // TODO: [Alpha] - is not used (and it seems that uint8_t is enough for this parameter)
```
Y-Less: So it isn't! The return should be:

```cpp
return vehicleColours[index] | alpha;
```

Y-Less:
Thanks, there are a few bad magic numbers in the code, thank you for looking at them.

temajm:
You can make the int index parameter of the carColourIndexToColour(int index, uint8_t alpha = 0xFF) function a uint8_t type and then the following check will not be needed (Or is it still important to use exactly int and check it for a range of 0 to 255?): 

image

SDK/include/values.hpp
temajm: I think this is unnecessary, and still the final index of the interval is useful (There are places where the interval is checked inclusive with the last index)

AmyrAhmady:
Please rebase your PR based on latest changes in repository if you are still willing to contribute to open.mp with these changes.
I'll have a look at them when it's ready

PR #623, opened 2023-01-30
Use `INVALID_VEHICLE_ID` for fail return of vehicles's functions

xunder-matth:


Alasnkz:
Why? They involve returning a player ID. Thus it should give an invalid player ID.

xunder-matth:
> Why? They involve returning a player ID. Thus it should give an invalid player ID.

That's a really good point. I don't know what I was thinking about, my bad.

PR #624, opened 2023-01-30
Validate `roll` in unoccupied sync

xunder-matth:


Server/Source/player_pool.hpp
```diff
@@ -1419,6 +1419,11 @@ struct PlayerPool final : public IPlayerPool, public NetworkEventHandler, public
 				return false;
 			}
 
+			if (std::fabs(1.0 - glm::length(unoccupiedSync.Roll)) >= 0.000001)
```
Southclaws: worth turning this magic number into a const and documenting the rationale behind the precision

NexiusTailer:
`direction` also must be validated, [see this](https://github.com/katursis/Pawn.RakNet/wiki/Unoccupied-sync-validity-checks).

There are two rules that the `roll` and `direction` vectors must comply with:
1. Range of values. `roll` and `direction` can be valid if they > -1.0 and < 1.0 (NaN values must be also discarded)
2. These vectors must be orthogonal to each other

So, the checks by the link above consider both things and cover all possible cases.

xunder-matth:
> `direction` also must be validated, [see this](https://github.com/katursis/Pawn.RakNet/wiki/Unoccupied-sync-validity-checks).

I validated `roll` because it can be abused to crash other players. I tried recreating crash with `direction` but I didn't get it to work. However looking at link you sent we should probably validate all data sent just to be sure it can't be exploited.

NexiusTailer:
> I validated `roll` because it can be abused to crash other players. I tried recreating crash with `direction` but I didn't get it to work.

Download any unoccupied ("quantum") crasher and check the basic ways of how they modified both direction and roll. Both of them can get any invalid values out of range (or even inside the range, but non-orthogonal) and they are all produce the same effects in the end for all streamed players (lite result is when the unoccupied vehicle just becomes invisible and more damage will occur if player are in spectating on this vehicle or surfing on it).

> However looking at link you sent we should probably validate all data sent just to be sure it can't be exploited.

I'm pretty sure that at least NaN/infinity cases cannot be achieved in omp server even now because iirc some filter for all float data was built-in somewhere while processing the sync. The checks that really needs to be added are actual ranges (-1.0 < roll/dir < 1.0) and orthogonal check for two arrays (vectors) to each other.

AGraber (commented):
Any reason for it to be `std::fabs(1.0 - glm::length(unoccupiedSync.Roll)) >= 0.000001` instead of `std::fabs(glm::length(unoccupiedSync.Roll)) > 1.0`? I'm rusty on all this stuff, but I just wanna know what's the deal behind that

AmyrAhmady (approved):

PR #648, opened 2023-03-18
Amir/changes

AmyrAhmady:
[return time in Move(Player)Object](https://github.com/openmultiplayer/open.mp/commit/3ee4d4e492628e4102fb0546b5785994eaa87d6f)
[lower priority for connect events in pawn comp](https://github.com/openmultiplayer/open.mp/commit/30f766517ad5d3781da4966f7e8bd51b1f459bad) (Because many player data are not initialized yet and people are using related pawn natives in OnPlayerConnect, like player object creation)
[port some security changes from samp 0.3.7](https://github.com/openmultiplayer/open.mp/commit/5518ab5b052766064cd92091078779a9094ab18e) (based on reversed samp server from 0.3.7)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -60,7 +60,7 @@ jobs:
       - name: Install latest conan
         run: |
           python -m pip install --upgrade pip
-          pip install conan
+          pip install -v "conan==1.57.0"
 
       - name: Generate build files
         run: |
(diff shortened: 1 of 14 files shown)


ksenonadv (approved):

IstuntmanI:
If I'm not wrong, the SA-MP Move(Player)Object functions aren't returning a float, but an int.

Issue #643, opened 2023-03-11
[BUG]printf and print


ghost:
"printf" and "print" functions cannot print Chinese characters, but it is possible to print Chinese characters in SA-MP (San Andreas Multiplayer).  Perhaps characters from other countries cannot be printed as well.

NexiusTailer:
Related to #637 #617 and #583

ghost:
Visual Studio Code
Files: Encoding: Simplified Chinese (GB2312)

**open.mp server version**: R2

**code:**
`main() {
	print("print中文测试!");
	printf("printf中文测试 %d", 1);
}`

**output:**
[2023-03-12T21:22:43+0800] [Info] print
[2023-03-12T21:22:43+0800] [Info] printf
[2023-03-12T21:22:43+0800] [Info] Legacy Network started on port 7777

AmyrAhmady:
Fixed in next build.

Issue #637, opened 2023-02-23
[BUG] Strings with cyrillic don't displayed with print(f)


NexiusTailer:
~~Packed~~ string with cyrillic don't displayed with both `print` and `printf` in the server console, while using non-formatted SendClientMessage or format (+ SendClientMessage or anything else in-game) it shows good.

Example script:
```pawn
#pragma pack true

main()
{
	print("Hi, I'm an English packed non-formatted string!");
	printf("Hi, I'm an English packed formatted string %d!", 2);
	print("Привет, я запакованная неотформатированная строка на русском!");
	printf("Привет, а я запакованная отформатированная строка на русском %d!", 2);
}
```

Output:
![image](https://user-images.githubusercontent.com/13169094/221046325-98956162-0643-4b2a-91da-b03956b3eb42.png)

UPD: Non-packed cyrillic strings shows the same behaviour, so, the problem is not related to packing itself.
UPD 2: Okay, it's definitely related to #583 and probably to #617 as well. I actually thought it was fixed after bt11...

Issue #617, opened 2023-01-16
While using Windows 1250 or 1252 encoding a bug occurs with texts that contains unicode characters


Mortyyka:
# Description

While using encoding **Windows 1250 or Windows 1252** this bug occours.

# How to re-produce this bug

Compile your script using **Windows 1250 or Windows 1252**, print out a text that contains unicode charaters in this example it was: **árvíztűrő tükörfúrógép**.

# Relevant log output

Console Output (Using Windows 1250 or Windows 1252)
[Image #1](https://asked.hu/XzkovGSD)

The same thing but in a text:

```
[Info] [OMP - Test]: Testing unicode characters:
[Info] [OMP - Test]: Testing unicode characters:
[Info] [OMP - Test]: Testing unicode characters:
[Info] [OMP - Test]: Testing unicode characters:
[Info] [OMP - Test]: Testing unicode characters:
[Info] [OMP - Test]: Testing unicode characters:
[Info] [OMP - Test]: Testing unicode characters:
[Info] [OMP - Test]: Testing unicode characters:
[Info] [OMP - Test]: Testing unicode characters:
[Info] [OMP - Test]: Testing unicode characters:
[Info] [OMP - Test]: Testing unicode characters:
[Info] [OMP - Test]: Testing unicode characters:
[Info] [OMP - Test]: Testing unicode characters:
[Info] [OMP - Test]: Testing unicode characters:
[Info] [OMP - Test]: Testing unicode characters:
[Info] [OMP - Test]: Testing unicode characters:
[Info] [OMP - Test]: Testing unicode characters:
[Info] [OMP - Test]: Testing unicode characters:
[Info] [OMP - Test]: Testing unicode characters:
[Info] [OMP - Test]: Testing unicode characters:
```

Note that in the server log it shows the whole text, including the unicode characters.

This is how the original message should look like:

```
[Info] [OMP - Test]: Testing unicode characters: árvíztűrő tükörfúrógép on OMP - RC2, text without unicode: arvizturo tukorfurogep
```

# open.mp server version

OMP RC2 - Latest Release

# Operating system or distribution

Windows 10

# Contact information

Mortyyka キング#1972

# Additional information

While using UTF-8 it works however, ingame the text becomes scuffed, example below.

```
[...]

Issue #583, opened 2022-12-27
[BUG]: Russian language support in the server console


MironGHW:
# Description

[
![image](https://user-images.githubusercontent.com/79435201/209658170-f24d8601-7832-4d46-aa5b-921bebd19bf1.png)
](url)

# open.mp server version

build11

# Operating system or distribution

windows10

# Contact information

M1RoN#6464

NexiusTailer:
> Set the console text encoding to the desired encoding or just compile your script with UTF-8 encoding.

Any pawn script which contains cyrillic **must** be compiled in Windows-1251. Any other encoding including UTF-8 will produce ?????? or other artifacts in the running server. The problem is in omp server console and the way how it works with pawn script's outputs.

PR #656, opened 2023-05-15
Fixed incorrect locale saving in Core::vlogLnInternal().

dev-karpov0:
There was a bug in function `Core::vlogLnInternal(...)` related to incorrect saving of the locale when `utf8=true`. It led to incorrect text output to the console after calling this function.

Reason: `std::setlocale()`, if success, returns a pointer to a new locale, not an old.

To recreate a bug:
1) print some string from script using non-latin symbols e.g. Cyrillic (cp1251)
2) Do something that make `Core::vlogLnInternal(.utf8=true)` call. (e.g. Input invalid command into console)
3) Print string again

### DIFF
--- a/Server/Source/core_impl.hpp
+++ b/Server/Source/core_impl.hpp
@@ -1676,24 +1676,35 @@ public:
 
 	virtual void vlogLnInternal(LogLevel level, bool utf8, const char* fmt, va_list args)
 	{
-#ifdef BUILD_WINDOWS
-		_lock_locales();
-		UINT oldCP = 0;
-		const char* oldLocale = nullptr;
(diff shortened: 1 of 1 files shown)


Hual (approved):

ksenonadv (approved):

Hual:
This should fix #643 #637 #617 and #583

Y-Less:
Why was debug level filtering added as well via this?

dev-karpov0:
That condition was not added but moved higher (before the `_lock_locales()` call). Otherwise, if condition were true, `_unlock_locales()` would not be called.

Issue #647, opened 2023-03-14
INetworkQueryExtension::onAddRule needs to be called every onTick


japajoe:
As discussed on Discord, there appears to be a bug with the INetworkQueryExtension::addRule method. For example take this code

```cpp
for (INetwork *network : core->getNetworks())
{
    INetworkQueryExtension* query = queryExtension<INetworkQueryExtension>(network);

    if (query)
    {
        query->addRule("worldtime", "Monday 12:30");
    }
}   
```

The expected behavior is that you see the text **Monday 12:30** in the SA-MP client, but unless you call this method every onTick, it will not display/change at all. Initially I was calling the method in 1 seconds intervals but I noticed I couldn't see this value in the client. The solution was as I described earlier, to call it every onTick.

Edit: While trying to remove a rule with INetworkQueryExtension::removeRule I've noticed the same behavior, not sure whether this is intended or not.

ksenonadv:
addRule works properly.
The server is overwriting worldtime rule each time there's an update triggered by ICore::updateNetworks()

japajoe:
According to one of the devs on Discord it shouldn't work like this and he told me to open an issue. I don't have a dog in this race however but if this is the desired behavior then it should be documented.

Y-Less:
PR #661

AmyrAhmady:
Since Y_Less's PR is now closed, Fixed in PR: #683

japajoe:
Beautiful thanks.

PR #658, opened 2023-05-17
Remove shared buffer svprintf function and replace it with an AmxStringFormatter class

Hual:
Let people override server rules, fixing #647

### DIFF
--- a/Server/Components/LegacyNetwork/Query/query.cpp
+++ b/Server/Components/LegacyNetwork/Query/query.cpp
@@ -164,9 +164,9 @@ void Query::buildRulesBuffer()
 		writeToBuffer(output, rule.first.c_str(), offset, ruleNameLength);
 
 		// Write rule value
-		uint8_t ruleValueLength = static_cast<uint8_t>(rule.second.length());
+		uint8_t ruleValueLength = static_cast<uint8_t>(rule.second.first.length());
 		writeToBuffer(output, offset, ruleValueLength);
-		writeToBuffer(output, rule.second.c_str(), offset, ruleValueLength);
+		writeToBuffer(output, rule.second.first.c_str(), offset, ruleValueLength);
 	}
 }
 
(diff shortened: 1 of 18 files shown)


Y-Less:
The custom rules fix is fine, no issues there.

Replacing the shared string buffer with separate ones is also a good thing in theory, but I'm concerned about the effects of allocating and initialising a new 8192 element array in every native involving strings.  There's no way that two different pawn functions with this formatting can be called at the same time, the re-entrant functions like `CallLocalFunction` aren't formatted so they could be left as a shared buffer - either just the old code or a pawn-component-wide `AmxStringFormatter` instance. Alternatively `query->addRule` already makes a full copy of the strings, so having the buffers separate shouldn't contribute directly to this fix and so could be moved to a separate PR.

Y-Less:
If the formatting were moved to a separate PR it would be good if the new buffer were expandable, removing the length checks entirely.  And then a shared buffer would slowly grow to the largest string ever seen anywhere in a script.

Hual:
> The custom rules fix is fine, no issues there.
> 
> Replacing the shared string buffer with separate ones is also a good thing in theory, but I'm concerned about the effects of allocating and initialising a new 8192 element array in every native involving strings. There's no way that two different pawn functions with this formatting can be called at the same time, the re-entrant functions like `CallLocalFunction` aren't formatted so they could be left as a shared buffer - either just the old code or a pawn-component-wide `AmxStringFormatter` instance. Alternatively `query->addRule` already makes a full copy of the strings, so having the buffers separate shouldn't contribute directly to this fix and so could be moved to a separate PR.

Neither allocation nor initialization happens with static arrays, it would perform exactly the same

Y-Less:
Ok you're right about the initialisation, but the class is over 8kb.  It isn't dynamic allocation, but it's still a lot of stack every time.  However, without the initialisation overhead this is much less of a problem.

Y-Less (approved):

PR #659, opened 2023-05-18
Fix MoveObject & MovePlayerObject return type. Fix ping packet spam.

ksenonadv:


Server/Components/Pawn/Scripting/Object/Natives.cpp
```diff
@@ -110,7 +110,7 @@ SCRIPT_API(IsValidObject, bool(IObject* object))
 	return object != nullptr;
 }
 
-SCRIPT_API(MoveObject, float(IObject& object, Vector3 position, float speed, Vector3 rotation))
+SCRIPT_API(MoveObject, int(IObject& object, Vector3 position, float speed, Vector3 rotation))
```
Y-Less, line 113: I was surprised it was a float.

Y-Less (approved):

PR #660, opened 2023-05-20
Fixed possibility to load the same script more than once.

dev-karpov0:
User could load the same script more than once the following ways:

1) load script with and without extension
```
loadfs test_script
loadfs test_script.amx
```

2) load script by command `loadscript` with rel. path to script that contains different slashes, e.g.
```
loadscript filterscripts/test_script
loadscript filterscripts\test_script
```

Added new function `utils::CanonicaliseScriptName()` that canonized path to script and adds an extension to the name of the script if it is missing. This ensures that the `PawnScript::name_` will always contain the name + extension of script.

### DIFF
--- a/Server/Components/Pawn/utils.hpp
+++ b/Server/Components/Pawn/utils.hpp
@@ -804,21 +804,6 @@ inline bool Canonicalise(std::string path, std::string& result)
 	return ret;
 }
 
-inline bool CanonicaliseScriptName(const std::string& name, std::string& result)
-{
-	if (utils::Canonicalise(name, result))
-	{
-		// if the user just supplied a script name, add the extension
-		// otherwise, don't, as they may have supplied a full abs/rel path.
-		if (!utils::endsWith(result, ".amx"))
-		{
-			result.append(".amx");
-		}
-		return true;
-	}
-	return false;
-}
-
 inline bool GetCurrentWorkingDirectory(std::string& result)
 {
 	result.clear();
@@ -866,6 +851,22 @@ inline bool GetCurrentWorkingDirectory(std::string& result)
 }
 
 #endif
+	
+inline bool CanonicaliseScriptName(const std::string& name, std::string& result)
+{
+	if (utils::Canonicalise(name, result))
+	{
+		// if the user just supplied a script name, add the extension
+		// otherwise, don't, as they may have supplied a full abs/rel path.
+		if (!utils::endsWith(result, ".amx"))
+		{
+			result.append(".amx");
+		}
+		return true;
+	}
+	return false;
+}
+	
 }
 
 static const FlatHashMap<String, String> DeprecatedNatives {
--- a/Server/Components/Pawn/utils.hpp
+++ b/Server/Components/Pawn/utils.hpp
@@ -849,24 +849,30 @@ inline bool GetCurrentWorkingDirectory(std::string& result)
 	}
 	return ret;
 }
-
 #endif
-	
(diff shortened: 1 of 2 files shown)


Y-Less:
You can load scripts twice, and I've seen this exploited to reload a script by loading the second one first, sharing data between them, then closing the original.  I don't know if those weird exploits are worth preserving this behaviour for, but it is at least worth discussing.  Having said that, I've also seen this exploit done using relative paths like `loadfs ..\filterscripts\x` vs `loadfs x`, 
and we already fixed that so...

dev-karpov0:
> I've also seen this exploit done using relative paths like `loadfs ..\filterscripts\x` vs `loadfs x`, and we already fixed that so...

No, it's not fixed (at least in RC2 release).

>  I've seen this exploited to reload a script by loading the second one first, sharing data between them, then closing the original

It's quite useful way to save data while restarting scripts, but the way to reload scripts shouldn't be so weird.

I made this fix because I was getting errors when entering commands `reloadfs` and `reloadscript` (as if the scripts were not loaded). It was confusing until I figured out what was going on.

My config file looks like this:
```
"side_scripts": [
    "filterscripts\\script1.amx", 
    "filterscripts\\script2.amx", 
]
```
Entering `reloadfs script1` results an error.
In Windows it's impossible to reload script by `reloadfs` if you use `\\` slash in config, only by `reloadscript` like that: `reloadscript filterscripts\script1.amx`. If script name (in config) given with extension you must also specify that ext. in command. Otherwise it results an error.

Y-Less:
> No, it's not fixed (at least in RC2 release).

Oh, well then...

> It's quite useful way to save data while restarting scripts, but the way to reload scripts shouldn't be so weird.

In that case I'd agree.  Fix this so `.amx` or not makes no difference and keep the `../` trick.  Not reloading scripts because you did or didn't use the extension is just weird as you say, but explicitly bypassing the system by using `../` is a deliberate act we should allow IMHO.

dev-karpov0:
The last commit is ready for review. 
FIxed `/`, `\` in path and now specifying (or not) the extension when (re)loading script not results an error if script was loaded with (or without) it.

Server/Components/Pawn/utils.hpp
```diff
@@ -849,8 +849,30 @@ inline bool GetCurrentWorkingDirectory(std::string& result)
 	}
 	return ret;
 }
-
 #endif
+inline void NormalizeScriptName(std::string name, std::string& result)
+{
+#ifdef WIN32
```
Zorono: since open.mp can be compiled as 64 bit binary..
it is preferred to be:
```c++
#if defined(_WIN32) || defined(WIN32) || defined(__WIN32__) || defined(_WIN64) || defined(WIN64) || defined(__WIN64__)
```

dev-karpov0: Thank you!

Server/Components/Pawn/utils.hpp
```diff
@@ -852,7 +857,7 @@ inline bool GetCurrentWorkingDirectory(std::string& result)
 #endif
 inline void NormalizeScriptName(std::string name, std::string& result)
```
AmyrAhmady: One last thing in here, rename it to `Normalise` since the entire code base is based on UK English, then I'll do a full review, thanks

AmyrAhmady (dismissed):

Y-Less (approved):

PR #661, opened 2023-05-20
Fix custom rules getting overridden by automatic values every tick.

Y-Less:
This keeps a set of keys that have had a custom value given, and if a key is found in that set only a new custom value can be assigned to it - it can't be auto-assigned.  Deleting a variable is also customising it so this flag is set in that case as well, which gives the interesting property that we can find deleted keys by checking if they are customised but don't exist.

Also, adding queries used to use a templated function with variable arguments, but then immediately assumed the parameters were convertible to strings.  There was nothing to enforce this in the signature and none of the calls took advantage of this in any way so I just removed it and used two `StringView` parameters.

Also now adds a load of new functions to the query rules system:

* `ResetServerRule` - Used after setting a custom value to return the rule to an automatically set state.
* `IsCustomServerRule` - `true` for rules set by script, `false` for those set by the server.
* `IsRemovedServerRule` - `true` for deleted rules.
* `IsProtectedServerRule` - `true` for rules that can't be modified.
* `ProtectServerRule` - Mark a rule as protected.  Once set can never be unset (that would defeat the object of protection).
* `GetServerRule` - Return the string value of a rule.

### DIFF
--- a/SDK/include/network.hpp
+++ b/SDK/include/network.hpp
@@ -371,6 +371,30 @@ struct INetworkQueryExtension : public IExtension
 	virtual bool isValidRule(StringView rule) = 0;
 };
 
+static const UID QueryRulesExtension_UID = UID(0xBC2B6AB9E411D4FB);
+struct IQueryRulesExtension : public IExtension
+{
+	PROVIDE_EXT_UID(QueryRulesExtension_UID);
+
+	/// Check if a rule is customised.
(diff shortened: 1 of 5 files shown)


AmyrAhmady:
Going to close this in favor of #683 where it is handling custom rules in a better, simpler, and less complicated way (not saying this one is complicated and hard to understand, just seeing it as being over-engineered)
We can always go back and check this PR if it's needed.

PR #663, opened 2023-05-21
Ported several functions missed out from fixes.inc.

Y-Less:
Most of these were trivial as the data already existed in the server:

* `GetDefaultPlayerColour` - just a wrapper for a newly abstracted `PlayerPool::getDefaultColour` function.
* `GetVehicleSeats` - wraps `getVehiclePassengerSeats`.
* `VehicleCanHaveComponent` - wraps `isValidComponentForVehicleModel`.
* `IsValidAnimationLibrary` - wraps `animationLibraryValid`, including an optional second parameter (default `true`).
* `PlayerHasClockEnabled` - wraps `player.hasClock`.

Although the server also missed `IsMenuValid` it has `IsValidMenu` instead, though that doesn't fix the syntax pattern on the other `IsMenuXXX` functions.  I'll leave this fix to includes for now, we can maybe think about updating the server later.

### DIFF
--- a/Server/Components/Pawn/Scripting/Player/Natives.cpp
+++ b/Server/Components/Pawn/Scripting/Player/Natives.cpp
@@ -214,6 +214,12 @@ SCRIPT_API(GetPlayerColor, int(IPlayer& player))
 	return player.getColour().RGBA();
 }
 
+SCRIPT_API(GetDefaultPlayerColour, int(int pid))
+{
+	// The player doesn't need to be connected for this to work.
+	return PawnManager::Get()->players->getDefaultColour(pid).RGBA();
+}
+
 SCRIPT_API(GetPlayerDrunkLevel, int(IPlayer& player))
 {
 	return player.getDrunkLevel();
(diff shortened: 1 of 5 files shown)


Server/Components/Pawn/Scripting/Player/Natives.cpp
```diff
@@ -292,6 +298,11 @@ SCRIPT_API(TogglePlayerClock, bool(IPlayer& player, bool enable))
 	return true;
 }
 
+SCRIPT_API(PlayerHasClockEnabled, bool(IPlayer& player))
```
Zorono, line 301: i suggest that function name can be changed to **IsPlayerClockEnabled**

Zorono, line 301: i suggest that function name can be changed to **IsPlayerClockEnabled**

Y-Less: The names are lifted straight from old includes.

AmyrAhmady (approved):

Issue #665, opened 2023-05-26
Kick doesn't properly disconnect a player under x64 build


NexiusTailer:
Just kick an existing player with any method (would it be Kick function from your pawn script or the rcon command), the player will see "Server closed the connection" message in his game but there will be no any disconnect message in the server console about it, as well as this player won't be able to reconnect to the server. Seems like his connection somehow keeps alive or something is hanging there after kicking a player.

Tested on:
- The latest artifact from actions, but it's reproducible on any recent ones.

- Windows 64-bit server build (my machine is Windows 10 Pro 22H2).
  - Linux x64 build wasn't tested at this moment.

AmyrAhmady:
TEsted by @NexiusTailer himself, seems to be okay now in PR #671 , so it is fixed in next build.

Issue #655, opened 2023-04-29
SetVehicleSpawnInfo delay & interiorid


BeckzyBoi:
When used, purely as example:

`public OnGameModeInit()
{
    new vehid = AddStaticVehicleEx(400, 0.0, 0.0, 10.0, 0.0, 0, 0, 300);
    SetVehicleSpawnInfo(vehid, 401, 0.0, 0.0, 10.0, 0.0, 0, 0, 300, -2);
    SetVehicleToRespawn(vehid);
}`

The vehicle model is changed to 401, and the respawn delay is/should still be 300, while the interior ID should also remain the same; however the vehicle becomes invisible (not interior ID 0?).

When keeping the respawn delay the same using -2, the vehicle appears as it should:

`SetVehicleSpawnInfo(vehid, 401, 0.0, 0.0, 10.0, 0.0, 0, 0, -2, -2);`

Basically, it appears that using anything other than -2 for the delay, OR using the same delay by definition, then using -2 to keep the same interiorid, won't correctly set the interiorid.

I can do some further digging if required. Thanks for your time.

AmyrAhmady:
Confirmed to be fixed in next build by OP

Issue #642, opened 2023-03-11
Unoccupied vehicles cannot trigger the OnVehicleDeath callback.


ghost:
In older versions of OpenMP, when an unoccupied vehicle falls into water, it triggers OnVehicleDeath, but in RC2, it no longer works.

AmyrAhmady:
Fixed in next build.

Issue #635, opened 2023-02-23
Give access to some extra sync data in new built-in natives


NexiusTailer:
Would be great to add a native like:
* GetVehicleParamsForPlayer. Just the getter for already existed SetVehicleParamsForPlayer
* GetPlayerAnimationFlags or GetPlayerAnimFlags. The information about anim flags is stored in [onfoot synchronization](https://github.com/katursis/Pawn.RakNet/wiki/OnFootSync)
* IsPlayerInDriveByMode. The drive-by flag is stored in [passenger synchronization](https://github.com/katursis/Pawn.RakNet/wiki/PassengerSync)

Others seems to be implemented when porting YSF stuff, but these two were mentioned by me on CBT in [this issue](https://github.com/openmultiplayer/server-beta-test/issues/20) among other functions from YSF (and probably were missed just because I didn't notice it's not from a plugin itself, but still useful to be added by default).

AmyrAhmady:
Except `GetVehicleParamsForPlayer`, the rest are done, and tested by Nexius himself
the remaining native will be added alongside other YSF native additions, closing this issue in favor of YSF todo list one

PR #671, opened 2023-06-08
Bunch of improvements, bug fixes and additions

AmyrAhmady:
#### Enhancements: 
- Add new natives #635 #592
```
native GetVehicleOccupant(vehicleid, seatid);
native GetVehicleMaxPassengers(modelid);
native CountVehicleOccupants(vehicleid);

native GetPlayerAnimFlags(playerid);
native GetPlayerAnimationFlags(playerid);
native IsPlayerInDriveByMode(playerid);
native IsPlayerCuffed(playerid);
```
- Ability to pass custom config path to open.mp as a CLI argument to read from #592
- Add `logging.file` config option to create/open/write into desired log file #133

#### Fixes:
- Fix seat ID issues related to incorrect passenger sync structure
- Improve passenger sync with new RE'd info (driveBy is actually one bit, the other remaining bit is *cuffed* status for passenger)

### Other changes made for issues:
- Print log with debug level in needed places #634 

---------------------
### Changes taken from forks
These changes are cherry-pick'd from https://github.com/Hual/open.mp (not all, some are left out)
issues which are allegedly resolved: 

- [x] #642
- [x] #610
- [x] #584
- [x] #625
- [x] #626
- [x] #628
- [x] #591
- [x] #655
- [x] #643
- [x] #637
- [x] #617
- [x] #583
- [x] #601
- [x] #665
- [x] #613
- [x]  https://github.com/openmultiplayer/RakNet/issues/16

All issue creators will be asked to test them out and give us feedback.

### DIFF
--- a/Server/Components/Pawn/Scripting/Player/Natives.cpp
+++ b/Server/Components/Pawn/Scripting/Player/Natives.cpp
@@ -669,28 +669,19 @@ SCRIPT_API(GetPlayerSpecialAction, int(IPlayer& player))
 	return player.getAction();
 }
 
-SCRIPT_API(GetPlayerVehicleID, int(IPlayer& player))
+SCRIPT_API_FAILRET(GetPlayerVehicleID, INVALID_PLAYER_ID, int(IPlayerVehicleData& data))
 {
-	IPlayerVehicleData* data = queryExtension<IPlayerVehicleData>(player);
-	if (data)
+	IVehicle* vehicle = data.getVehicle();
+	if (vehicle)
 	{
-		IVehicle* vehicle = data->getVehicle();
-		if (vehicle)
-		{
-			return vehicle->getID();
-		}
+		return vehicle->getID();
 	}
 	return 0;
 }
 
-SCRIPT_API_FAILRET(GetPlayerVehicleSeat, SEAT_NONE, int(IPlayer& player))
+SCRIPT_API_FAILRET(GetPlayerVehicleSeat, SEAT_NONE, int(IPlayerVehicleData& data))
 {
-	IPlayerVehicleData* data = queryExtension<IPlayerVehicleData>(player);
-	if (data)
-	{
-		return data->getSeat();
-	}
-	return FailRet;
+	return data.getSeat();
 }
 
 SCRIPT_API(GetPlayerWeaponData, bool(IPlayer& player, int slot, int& weaponid, int& ammo))
@@ -817,32 +808,16 @@ SCRIPT_API_FAILRET(GetPlayerTargetActor, INVALID_PLAYER_ID, int(IPlayer& player)
 	return FailRet;
 }
 
-SCRIPT_API(IsPlayerInVehicle, bool(IPlayer& player, IVehicle& targetVehicle))
+SCRIPT_API(IsPlayerInVehicle, bool(IPlayerVehicleData& data, IVehicle& targetVehicle))
 {
-	IPlayerVehicleData* data = queryExtension<IPlayerVehicleData>(player);
-	if (data)
-	{
-		IVehicle* vehicle = data->getVehicle();
-		if (vehicle == &targetVehicle)
-		{
-			return true;
-		}
-	}
-	return false;
+	IVehicle* vehicle = data.getVehicle();
+	return vehicle == &targetVehicle;
 }
 
-SCRIPT_API(IsPlayerInAnyVehicle, bool(IPlayer& player))
+SCRIPT_API(IsPlayerInAnyVehicle, bool(IPlayerVehicleData& data))
 {
-	IPlayerVehicleData* data = queryExtension<IPlayerVehicleData>(player);
-	if (data)
-	{
-		IVehicle* vehicle = data->getVehicle();
-		if (vehicle)
-		{
-			return true;
-		}
-	}
-	return false;
+	IVehicle* vehicle = data.getVehicle();
+	return vehicle != nullptr;
 }
 
 SCRIPT_API(IsPlayerInRangeOfPoint, bool(IPlayer& player, float range, Vector3 position))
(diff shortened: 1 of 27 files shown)


Server/Components/Pawn/Scripting/Player/Natives.cpp
```diff
@@ -1241,3 +1241,44 @@ SCRIPT_API(IsPlayerUsingOfficialClient, int(IPlayer& player))
 {
 	return player.isUsingOfficialClient();
 }
+
+SCRIPT_API(GetPlayerAnimFlags, int(IPlayer& player))
+{
+	PlayerAnimationData data = player.getAnimationData();
```
Hual, line 1222: For the future: all the animation string arrays etc should go in the Impl namespace and the Impl folder

Server/Components/Pawn/Scripting/Player/Natives.cpp
```diff
@@ -1241,3 +1241,44 @@ SCRIPT_API(IsPlayerUsingOfficialClient, int(IPlayer& player))
...
+}
+
+SCRIPT_API(GetPlayerAnimationFlags, int(IPlayer& player))
+{
+	return openmp_scripting::GetPlayerAnimFlags(player);
+}
+
+SCRIPT_API(IsPlayerInDriveByMode, bool(IPlayer& player))
+{
+	IPlayerVehicleData* data = queryExtension<IPlayerVehicleData>(player);
```
Hual: Better to add IPlayerVehicleData to the player pool params and use that in SCRIPT_API directly, maybe add other player data classes too

Server/Components/Pawn/Scripting/Vehicle/Natives.cpp
```diff
@@ -646,3 +646,50 @@ SCRIPT_API(GetVehicleMatrix, bool(IVehicle& vehicle, Vector3& right, Vector3& up
 	at = mat[2];
 	return true;
 }
+
+SCRIPT_API_FAILRET(GetVehicleOccupant, INVALID_PLAYER_ID, int(IVehicle& vehicle, int seat))
+{
+	IPlayer* driver = vehicle.getDriver();
+	const FlatHashSet<IPlayer*>& passengers = vehicle.getPassengers();
```
Hual: This should be moved to just before using passengers to avoid the virtual call

Server/Components/Pawn/Scripting/Vehicle/Natives.cpp
```diff
@@ -646,3 +646,50 @@ SCRIPT_API(GetVehicleMatrix, bool(IVehicle& vehicle, Vector3& right, Vector3& up
...
+	// Looking for driver
+	if (seat == 0)
+	{
+		return driver == nullptr ? INVALID_PLAYER_ID : driver->getID();
+	}
+	// Looking for a passenger
+	else
+	{
+		for (auto& passenger : passengers)
```
Hual, line 640: Wouldn't it make sense to store passengers as a map with seat ID as key? Lookups would be much faster then

Server/Source/core_impl.hpp
```diff
@@ -281,12 +283,16 @@ class Config final : public IEarlyConfig
 	}
 
 public:
-	Config(ICore& core, bool defaultsOnly = false)
+	Config(ICore& core, bool defaultsOnly = false, const cxxopts::ParseResult* cmd = nullptr)
```
Hual, line 286: Is there a reason cmd is a pointer here?

Hual (changes requested):

Server/Components/Pawn/Scripting/Vehicle/Natives.cpp
ksenonadv: > Wouldn't it make sense to store passengers as a map with seat ID as key? Lookups would be much faster then

What about some vehicles like bus which can have multiple players reporting same seat id.

Hual: > > Wouldn't it make sense to store passengers as a map with seat ID as key? Lookups would be much faster then
> 
> What about some vehicles like bus which can have multiple players reporting same seat id.

In that case the data could be a pair with seat ID

Hual: > > Wouldn't it make sense to store passengers as a map with seat ID as key? Lookups would be much faster then
> 
> What about some vehicles like bus which can have multiple players reporting same seat id.

Then technically it wouldn't fit all the passengers but it doesn't seem to need all of them for the current purposes (it would return the first one with that seat ID anyway)

Server/Source/core_impl.hpp
AmyrAhmady: Making it optional with nullptr as default value when Config class is used without any cmd option available to pass

Server/Components/Pawn/Scripting/Vehicle/Natives.cpp
AmyrAhmady: Using a map of passengers means doing lookup every tick per every received passenger sync, then decide whether to leave it, or do insertion or removal

Doing lookups and then possibly two other operation for every passenger sync VS having a pawn native and letting users decide how frequent they want it to be, I'd choose the second.

@Hual @ksenonadv

Server/Source/core_impl.hpp
Hual: But config is only constructed by core which has a ParseResult&

AmyrAhmady: There's also this one
https://github.com/openmultiplayer/open.mp/blob/54e8fc72e8ff71a4d4a89cab34e9a244266dbed2/Server/Source/core_impl.hpp#L615

Hual: Resolved then

Server/Components/Pawn/Scripting/Vehicle/Natives.cpp
Hual: Resolved then

Server/Components/Pawn/Scripting/Player/Natives.cpp
AmyrAhmady: Done

Server/Components/Pawn/Scripting/Vehicle/Natives.cpp
AmyrAhmady: Done

Hual (approved):

ksenonadv (approved):

PR #685, opened 2023-06-21
Implement (Start|Stop)RecordingPlayerData v2

Hual:
Work on top of #567


Server/Components/Recordings/recordings_main.cpp
```diff
@@ -92,7 +92,7 @@ class RecordingsComponent final : public IRecordingsComponent, public PlayerConn
 			}
 
 			// Write on foot recording data
-			if (data->file_.good() && data->type_ == PlayerRecordingType_OnFoot)
+			if (data->type_ == PlayerRecordingType_Driver && data->file_.good())
```
WoutProvost: Should be `data->type_ == PlayerRecordingType_OnFoot` for the on foot recording data.

Server/Components/Recordings/recordings_main.cpp
```diff
@@ -148,24 +124,7 @@ class RecordingsComponent final : public IRecordingsComponent, public PlayerConn
 			{
 				uint32_t timeSinceRecordStart = duration_cast<Milliseconds>(Time::now() - data->start_).count();
```
WoutProvost: Nitpicking, but if you make it `const` for the on foot code, it can become `const` here as well.

Zorono (approved):

WoutProvost:
I don't know enough about the NetworkBitStream related changes in [1670ea0](https://github.com/openmultiplayer/open.mp/pull/685/commits/1670ea085227835492eaae419528c550f7bd0d0f), so I'll leave that to @AmyrAhmady. Otherwise it looks good

AmyrAhmady:
@WoutProvost IS it possible for you to do a quick test and see if everything works fine?
record a simple NPC with both onfoot and driver modes and test it

WoutProvost:
> @WoutProvost IS it possible for you to do a quick test and see if everything works fine? record a simple NPC with both onfoot and driver modes and test it

@AmyrAhmady Finally gotten around to doing some tests. Data is successfully being recorded for both onfoot and driver modes respectively. However, when trying to analyze the recording files with my [tool](https://github.com/WoutProvost/samp-rec-to-csv), I received these errors:
```
./rec2csv: Source '../onfoot.rec' is corrupt.
./rec2csv: Source '../driver.rec' is corrupt.
```

My tool is specifically designed to error on file sizes that have a different amount of bytes than they should have. For the onfoot recording, this is equivalent to `(size-8) % 72 != 0` and for the driver recording, this is equivalent to `(size-8) % 67 != 0`, with the `8` denoting the header block and the `72` and `67` denoting the onfoot and driver blocks respectively.

In other words, this change is unfortunately not an exact match with how SA:MP does it, whereas my [implementation](https://github.com/openmultiplayer/open.mp/pull/567) is.

WoutProvost (changes requested):
This change broke backwards compatibility of this feature with SA:MP.

Y-Less:
My guess for that would be struct padding:

> I don't know enough about the NetworkBitStream related changes in [1670ea0](https://github.com/openmultiplayer/open.mp/pull/685/commits/1670ea085227835492eaae419528c550f7bd0d0f), so I'll leave that to @AmyrAhmady. Otherwise it looks good

This just dumps the packet directly to the file, rather than parsing all the data first then writing the components.  One thing I noticed is that the original code saves the health and armour as `uint8_t`, rather than `float` as they are in the `PlayerFootSync` packet.

Y-Less:
So I suspect this part needs reverting:

https://github.com/openmultiplayer/open.mp/pull/685/commits/1670ea085227835492eaae419528c550f7bd0d0f#diff-3ae0b742a2dd12c25e17048c1c17a9d06766574ceda694b905fabf8cc8c3e071L81-L171

ksenonadv:
> My guess for that would be struct padding:
> 
> > I don't know enough about the NetworkBitStream related changes in [1670ea0](https://github.com/openmultiplayer/open.mp/pull/685/commits/1670ea085227835492eaae419528c550f7bd0d0f), so I'll leave that to @AmyrAhmady. Otherwise it looks good
> 
> This just dumps the packet directly to the file, rather than parsing all the data first then writing the components. One thing I noticed is that the original code saves the health and armour as `uint8_t`, rather than `float` as they are in the `PlayerFootSync` packet.


Actually it's more subtle than that.
Bitstream contains the packet id which is not supposed to be saved.
This should work:

```cpp
data->file_.write(reinterpret_cast<const char*>(bs.GetData()[1]), bs.GetNumberOfBytesUsed() - 1);
```

WoutProvost:
I really don't know why this PR is still open and you're all still commenting here. My [PR](https://github.com/openmultiplayer/open.mp/pull/567), which was created 6 months earlier, was proven to work, whereas this one was proven to not work. We could've merged mine 8 months ago and people would've been able to use this feature already.

I really don't see what we're waiting for.

Y-Less:
Cherry-picked part of this, with the original file writing code.

PR #698, opened 2023-08-01
Add artwork models's web server port option to config file

Abyss-z94:
To support some old TCP plugin like Incognito's audio plugin (The server plugin allows creating TCP servers on different ports but the client can only connect to the server through TCP port 7777).

It gonna use default port is 7776 by default to create web server instead of 7777.


Server/Components/CustomModels/models.cpp
```diff
@@ -514,7 +521,7 @@ class CustomModelsComponent final : public ICustomModelsComponent, public Player
 			return;
 		}
 
-		webServer = new WebServer(core, modelsPath, core->getConfig().getString("network.bind"), *core->getConfig().getInt("network.port"), core->getConfig().getString("network.public_addr"), httpThreads);
+		webServer = new WebServer(core, modelsPath, core->getConfig().getString("network.bind"), *core->getConfig().getInt("artwork.port"), core->getConfig().getString("network.public_addr"), httpThreads);
```
Hual, line 524: nullptr check here since unlike getString() getInt() returns a pointer

Hual (changes requested):

Y-Less:
I put the default port back to 7777.  I understand why you used 7776 to avoid conflicts, but that would be a breaking change from existing server deployments.  So we need to leave that as-is, and let anyone who wants 7776 to configure it themselves.

PR #710, opened 2023-09-02
Checking abi

Y-Less:
I've merged a few things, so this isn't a true PR, just wanted to see the ABI tool output before pushing to master and closing the PRs.


Y-Less:
Yes I thought that was the case.  It's the `PawnLookup` struct, which isn't used in the public API, but is in the generic includes because it can be used by both the pawn component and plugins internally.  Thus, can be ignored.

PR #747, opened 2023-10-19
Correctly retrieve `modelid` and material colors in `GetObjectMaterial(Text)`

xunder-matth:


Vince0789 (approved):

Server/Components/Pawn/Scripting/Object/Natives.cpp
```diff
@@ -282,6 +282,7 @@ SCRIPT_API(GetObjectMaterial, bool(IObject& object, int materialIndex, int& mode
 	bool result = object.getMaterialData(materialIndex, data);
 	if (result)
 	{
+		modelid = data->model;
 		textureLibrary = data->textOrTXD;
 		textureName = data->fontOrTexture;
 		materialColour = data->materialColour.RGBA();
```
Vince0789: Not directly related to this PR but this should probably return ARGB to be consistent with `SetObjectMaterial`.

xunder-matth: That's also true for font and background colors in `GetObjectMaterialText`.

AmyrAhmady (approved):

PR #753, opened 2023-10-22
Fix move object rotation when optional args are not set

### LINKED ISSUE #749: [Behavioural difference] `Move(Player)Object` rotation
The rotation parameters for `Move(Player)Object` functions are optional and are set to `-1000.0` by default. In SA-MP server, any default value given is internally converted to the actual object rotation, and then the server sends the MoveObject RPC to the client with the correct rotation values. In open.mp server, the `-1000.0` value is not changed to the real object rotation value, resulting in the client receiving an RPC with incorrect rotation parameters.

ksenonadv:
Fixes #749

### DIFF
--- a/Server/Components/Objects/object.hpp
+++ b/Server/Components/Objects/object.hpp
@@ -213,16 +213,30 @@ protected:
 		moving_ = true;
 		moveData_ = data;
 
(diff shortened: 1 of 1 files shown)


xunder-matth:
There will be cases where you want to keep only 1 or 2 axes the same, so you will pass `_` (or `-1000.0`). However, your pull request only accounts for keeping all three parameters as default.

AmyrAhmady (approved):

PR #788, opened 2023-11-27
Fixed GetPlayerVehicleID returning INVALID_VEHICLE_ID.

kalixis:


Hual:
What does this fix exactly?

kalixis:
On the server, GetPlayerVehicleID returns INVALID_VEHICLE_ID, although on the website https://www.open.mp/docs/scripting/functions/GetPlayerVehicleID, it is stated to return 'ID of the vehicle or 0 if not in a vehicle.'

Or is it not fixed this way?

BeckzyBoi:
I could've sworn Amir fixed this earlier this year, but at another glance it seems to still be the same:

https://github.com/openmultiplayer/open.mp/commit/50876f92be7c418ff0eaa70cde6b49ea96502774

However I use GetPlayerVehicleID a lot and check if it equals, or doesn't equal, zero and I've had no issues at all. Strange..

Edit: Never mind, I still have it patched with GetPlayerVehicleID2:

`GetPlayerVehicleID2(playerid)
{
	new vid = GetPlayerVehicleID(playerid);
	if (vid == INVALID_VEHICLE_ID) return 0;
	return vid;
}`

So yeah I guess this fix is actually relevant.

PerseweratywneImponderabilia:
I'm pretty sure it makes more sense to return `INVALID_VEHICLE_ID` instead of `0`.
If samp used to return the value 0, it's definitely a mistake and the difference between samp and open.mp should be noted in the wiki whilst keeping the existing behavior here.

AmyrAhmady:
This break a lot of scripts depending on it, while it's good to stay consistent, we still try to stay as compatible as possible to 20 years oof samp scripts

Also documentation is wrong (or not complete to be precise). If playerid is not valid, it must return 0, and if player's vehicle is not valid, it must return INVALID_VEHICLE_ID. So you just have to reverse how original returns worked.

Server/Components/Pawn/Scripting/Player/Natives.cpp
```diff
@@ -665,14 +665,14 @@ SCRIPT_API(GetPlayerSpecialAction, int(IPlayer& player))
...
-SCRIPT_API_FAILRET(GetPlayerVehicleID, INVALID_VEHICLE_ID, int(IPlayerVehicleData& data))
+SCRIPT_API_FAILRET(GetPlayerVehicleID, 0, int(IPlayerVehicleData& data))
 {
 	IVehicle* vehicle = data.getVehicle();
 	if (vehicle)
 	{
 		return vehicle->getID();
 	}
-	return 0;
+	return FailRet;
```
Hual, line 675: return INVALID_VEHICLE_ID here

Hual (changes requested):

Server/Components/Pawn/Scripting/Player/Natives.cpp
kalixis: It's strange, but if you input an invalid player ID (a player who is not on the server) into GetPlayerVehicleID, it will return 65535 (INVALID_PLAYER_ID).

And if no player is in the vehicle, it will return 0.

PR #798, opened 2023-12-03
New way of handling ParamCast errors instead of throwing exceptions.

AmyrAhmady:
Well, this was discussed way too many times in our internal chats, throwing exceptions for simply failing casting parameters can cost us performance and it can be really expensive sometimes when you have no idea it's actually even happening and the native is being used rapidly.


Server/Components/Pawn/Scripting/Types.hpp
```diff
@@ -23,7 +23,7 @@ template <>
 class ParamCast<PawnScript&>
 {
 public:
-	ParamCast(AMX* amx, cell* params, int idx)
+	ParamCast(AMX* amx, cell* params, int idx, int& error = ParamCastErrorCode_None)
```
Hual: int& shouldn't have a default value

Hual (changes requested):

Server/Components/Pawn/Scripting/Types.hpp
AmyrAhmady: Entirely changed how it works now, no more ref args, handled internally in each ParamCast

Hual: There were 3 error values before but now it's just a bool, what happened to the third one?

AmyrAhmady: Not needed anymore because the old ones weren't used, only there for future if it would be ever needed but since now are error handlings actually happen in every ParamCast class, user/dev can make their own version of it (like printing something in case of failure), and return true at the end in ParamCast::Error().

We (pawn-native) only want to know if it failed or not, just to stop script native execution

SDK/include/Server/Components/Pawn/pawn_natives.hpp
```diff
@@ -20,486 +20,513 @@ using OutputOnlyString = std::variant<bool, StringView, Impl::String>;
 /// Example with IPlayer from the players pool:
 /// Using IPlayer& in a script function throws an exception if the player with the specified ID doesn't exist
 /// Using IPlayer* in a script function makes the IPlayer entry optional and is nullptr if the player with the specified ID doesn't exist
-#define POOL_PARAM(type, poolPtr)                            \
```
Hual, line 23: Can you put all this in the Impl namespace?

Hual (changes requested):

Hual (approved):

PR #799, opened 2023-12-04
Optimize batch packets & RPCs using RakNet list commands

Hual:


Server/Components/LegacyNetwork/legacy_network_impl.hpp
```diff
@@ -139,6 +139,85 @@ class RakNetLegacyNetwork final : public Network, public CoreEventHandler, publi
...
+					--playersSize;
+				}
+			}
+			if (playersSize == 1)
+			{
+				IPlayer* player = *players.begin();
+				if (player == exceptPeer)
+				{
+					return false;
+				}
```
Hual, line 159: should return sendPacket to the single player here

Server/Components/LegacyNetwork/legacy_network_impl.hpp
```diff
@@ -234,6 +313,86 @@ class RakNetLegacyNetwork final : public Network, public CoreEventHandler, publi
...
```
Hual, line 338: should return sendRPC to the single player here

PR #800, opened 2023-12-05
Update player_pool.hpp

akosd127:
Fix #792

### DIFF
--- a/Server/Source/player_pool.hpp
+++ b/Server/Source/player_pool.hpp
@@ -1663,8 +1663,8 @@ struct PlayerPool final : public IPlayerPool, public NetworkEventHandler, public
 				addressString.data(),
 				port,
 				player.poolID);
-			playerConnectDispatcher.dispatch(&PlayerConnectEventHandler::onIncomingConnection, peer, addressString, port);
 		}
+		playerConnectDispatcher.dispatch(&PlayerConnectEventHandler::onIncomingConnection, peer, addressString, port);
 		// Don't process player, about to be disconnected
 		if (player.kicked_)
 		{
--- a/Server/Source/player_pool.hpp
+++ b/Server/Source/player_pool.hpp
@@ -1655,14 +1655,16 @@ struct PlayerPool final : public IPlayerPool, public NetworkEventHandler, public
 		PeerAddress::AddressString addressString;
 		PeerAddress::ToString(player.netData_.networkID.address, addressString);
 		uint16_t port = player.netData_.networkID.port;
-		core.logLn(
-			LogLevel::Message,
-			"[connection] incoming connection: %s:%d id: %d",
-			addressString.data(),
-			port,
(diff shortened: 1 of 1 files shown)


Hual:
It would be nice if you would test your code before making a PR

Server/Source/player_pool.hpp
```diff
@@ -1655,14 +1655,16 @@ struct PlayerPool final : public IPlayerPool, public NetworkEventHandler, public
...
 		uint16_t port = player.netData_.networkID.port;
-		core.logLn(
-			LogLevel::Message,
-			"[connection] incoming connection: %s:%d id: %d",
-			addressString.data(),
-			port,
-			player.poolID);
-		playerConnectDispatcher.dispatch(&PlayerConnectEventHandler::onIncomingConnection, peer, addressString, port);
-
+		if (config.getBool("logging.log_connection_messages"))
```
Hual, line 1658: getBool returns a pointer that needs to be dereferenced

Server/Source/player_pool.hpp
```diff
@@ -1655,14 +1655,16 @@ struct PlayerPool final : public IPlayerPool, public NetworkEventHandler, public
...
+		{
+			core.logLn(
+				LogLevel::Message,
+				"[connection] incoming connection: %s:%d id: %d",
+				addressString.data(),
+				port,
+				player.poolID);
+			playerConnectDispatcher.dispatch(&PlayerConnectEventHandler::onIncomingConnection, peer, addressString, port);
```
Hual: calling the event shouldn't be in the if statement

Hual (changes requested):

Server/Source/player_pool.hpp
akosd127: I'm sorry if I did something wrong. I just copied your code thinking it was the right way. please fix it properly and add it to the master branch. Unfortunately, English is not my native language and I don't work in IT, I'm just an avid fan

akosd127 (commented):
check please

Server/Source/player_pool.hpp
```diff
@@ -1655,14 +1655,16 @@ struct PlayerPool final : public IPlayerPool, public NetworkEventHandler, public
 		PeerAddress::AddressString addressString;
 		PeerAddress::ToString(player.netData_.networkID.address, addressString);
 		uint16_t port = player.netData_.networkID.port;
-		core.logLn(
-			LogLevel::Message,
-			"[connection] incoming connection: %s:%d id: %d",
-			addressString.data(),
-			port,
-			player.poolID);
+		if (config.getBool("logging.log_connection_messages"))
```
Hual, line 1658: This should be *config.getBool(), same with the disconnect message; not sure how that one got in

Hual (changes requested):

Issue #797, opened 2023-12-02
weaponid isn't validated in basic synchronization types


NexiusTailer:
**Describe the bug**
A simple yet still unnoticed problem is that weaponid isn't validated as valid weapon id inside onfoot, driver and passenger synchronization sent by any player. This is pretty critical as GetPlayerWeapon is usually used as an array index.

**To Reproduce**
Spoof any type of sync (onfoot, driver or passenger) with invalid weaponid which is not in range of [0-18, 22-46] and it will be passed to GetPlayerWeapon which you can print right in OnPlayerUpdate to easily see when it reaches the server.

Important note is to send non-valid weaponid, but it should fit the datatype by its value (weapons can be 0-63 due to size of data in the bitstream argument, so send something more than 46 and less than 64).

**Expected behavior**
Is this case, SA-MP server just reset weaponid to 0 and the whole packet processes as usual.

**Screenshots and/or logs**
Could send my crashdetect log when I firstly discovered this issue ha-ha

**Release version**
v1.1.0.2612

BeckzyBoi:
Is this how some cheaters can use Minigun amongst other weapons in my server without it being detected (I'm using GetPlayerWeapon in OnPlayerUpdate to detect them)? I also wonder what GetPlayerWeaponData reports...

NexiusTailer:
No it's not about weapon hack working, it is a flaw related to the lack of validation for the **sane** values sent by clients updating their holding weapon. The worst result it can bring is tons of out-of-bounds runtime errors inside any functions where you used GetPlayerWeapon as an array index. Which is still a critical thing, isn't it?

> I also wonder what GetPlayerWeaponData reports...

GetPlayerWeaponData represent weapons update sync i.e. all weapons and ammo that the player has in every slot (while GetPlayerWeapon is just a "marker" which weapon from all those is in hands right now).

PR #804, opened 2023-12-09
Weapon validity check in foot, passenger, and vehicle sync, and Give/TakeDamage events

### LINKED ISSUE #600: Validation in OnPlayerGiveTakeDamage
https://github.com/openmultiplayer/open.mp/blob/dcd0190a91e67c99217e8ed3254baeb86733a0e4/Server/Source/player_pool.hpp#L215

https://github.com/openmultiplayer/open.mp/blob/dcd0190a91e67c99217e8ed3254baeb86733a0e4/Server/Source/player_pool.hpp#L249

These lines should have validate weapon IDs under TakeDamage and GiveDamage parts of the callback, but SA-MP client can send OnPlayerTakeDamage with a wider range of weapon IDs than in OnPlayerGiveDamage: damage can be taken by **reason**, while can be given by **weapon**.

The valid reasons for OnPlayerTakeDamage are:
0..18, 22..46, 49..54 (including 52, it's half-valid despite the samp docs says)

The valid weapon IDs for OnPlayerGiveDamage are:
0..18, 22..46

AmyrAhmady:
Fixes #600 and #797

### DIFF
--- a/Server/Source/player_pool.hpp
+++ b/Server/Source/player_pool.hpp
@@ -213,7 +213,17 @@ struct PlayerPool final : public IPlayerPool, public NetworkEventHandler, public
 					{
 						return false;
 					}
-					if (!from->areWeaponsAllowed() && 0 < onPlayerGiveTakeDamageRPC.WeaponID && onPlayerGiveTakeDamageRPC.WeaponID <= 47)
+
+					auto slot = WeaponSlotData(onPlayerGiveTakeDamageRPC.WeaponID).slot();
+					if (slot == INVALID_WEAPON_SLOT)
+					{
+						if (onPlayerGiveTakeDamageRPC.WeaponID < 49 || onPlayerGiveTakeDamageRPC.WeaponID > 54)
+						{
+							return false;
+						}
+					}
+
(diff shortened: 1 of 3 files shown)


Server/Source/player_pool.hpp
```diff
@@ -213,7 +213,17 @@ struct PlayerPool final : public IPlayerPool, public NetworkEventHandler, public
 					{
 						return false;
 					}
-					if (!from->areWeaponsAllowed() && 0 < onPlayerGiveTakeDamageRPC.WeaponID && onPlayerGiveTakeDamageRPC.WeaponID <= 47)
+
+					auto slot = WeaponSlotData(onPlayerGiveTakeDamageRPC.WeaponID).slot();
+					if (slot == INVALID_WEAPON_SLOT)
+					{
+						if (onPlayerGiveTakeDamageRPC.WeaponID < 49 || onPlayerGiveTakeDamageRPC.WeaponID > 54)
```
Hual: What does this check do? It seems to skip some weapon IDs that are only valid for death reasons but then it doesn't skip Connect, Disconnect, Suicide?

NexiusTailer: > What does this check do? It seems to skip some weapon IDs that are only valid for death reasons but then it doesn't skip Connect, Disconnect, Suicide?

It's actually damage reasons and they are not the same with death reasons just because of 200+ IDs (connect, disconnect, suicide).

ksenonadv (approved):

PR #819, opened 2023-12-28
Fix amx_Allot trying to access invalid memory when data size is higher than amx STK size

AmyrAhmady:
This PR prevents server from crashing when `amx_Allot` wrongfully returns `AMX_ERR_NONE` and servers thinks it was successful when data size is higher than STK size, by simply returning `AMX_ERR_MEMORY` on when that situation happens and also printing a message about it, while hinting the user what they can do to fix it in their pawn code.

Example:
```log
[2023-12-28T19:13:21+0330] [Error] Unable to find enough memory for your data.
[2023-12-28T19:13:21+0330] [Error] Size: 4195184 bytes, Available space: 17256 bytes, Need extra size: 4177928 bytes
[2023-12-28T19:13:21+0330] [Error] You can increase your available memory size by using `#pragma dynamic 1048796`.
[2023-12-28T19:13:21+0330] [Error] Out of memory
```

### DIFF
--- a/Server/Components/Pawn/Script/Script.cpp
+++ b/Server/Components/Pawn/Script/Script.cpp
@@ -12,7 +12,6 @@
 */
 
 #define _Static_assert static_assert
-#define STKMARGIN ((cell)(16 * sizeof(cell))) // from amx.c
 
 #include <assert.h>
 #include <stdarg.h>
--- a/Server/Components/Pawn/Script/Script.cpp
+++ b/Server/Components/Pawn/Script/Script.cpp
@@ -117,114 +117,6 @@ PawnScript::~PawnScript()
 	tryLoad("");
 }
 
-int AMXAPI amx_NumPublics(AMX* amx, int* number)
-{
-	AMX_HEADER* hdr = (AMX_HEADER*)amx->base;
-	assert(hdr != NULL);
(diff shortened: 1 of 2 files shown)


Server/Components/Pawn/Script/Script.cpp
```diff
@@ -453,3 +454,46 @@ int AMXAPI amx_Register(AMX* amx, const AMX_NATIVE_INFO* list, int number)
 		amx->flags |= AMX_FLAG_NTVREG;
 	return err;
 }
+
+int AMXAPI amx_Allot(AMX* amx, int cells, cell* amx_addr, cell** phys_addr)
```
Hual: See the amx_FindPublic comments and make sure these functions aren't being hooked by common plugins too

Server/Components/Pawn/Script/Script.cpp
```diff
@@ -12,6 +12,7 @@
 */
 
 #define _Static_assert static_assert
+#define STKMARGIN ((cell)(16 * sizeof(cell))) // from amx.c
```
Hual: Wouldn't this be better exposed through a function or common header so if it's ever changed it's consistent?

Hual (changes requested):

Hual (approved):

PR #820, opened 2023-12-29
Remove preconnect packet handler

AmyrAhmady:


Zorono:
excuse me
please can you explain this PR ??

AmyrAhmady:
server can receive packets before player instance being initialized, so open.mp was keeping "pre-connect" packets in a map to process them later when player is fully initialized. 
This is a very very very rare case that the only time I know it happens is when a user uses SAMP Android. It sends a client config data packet before connect packet to servers, so I was aiming to fix that by letting server process that config packet after successfully connecting and initializing the player instance, but this way of doing it caused more issues than fixing some issues that shouldn't have even existed in the first place.

Hual (approved):

PR #822, opened 2023-12-31
52 isn't a valid damage reason

Y-Less:
Excludes `52` being accepted as a valid damage reason, where previously `49 <= x <= 54` accepted it.

### DIFF
--- a/SDK/include/utils.hpp
+++ b/SDK/include/utils.hpp
@@ -18,13 +18,15 @@ inline StringView trim(StringView view)
 
 inline bool IsWeaponForTakenDamageValid(int weapon)
 {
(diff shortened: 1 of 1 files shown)


AmyrAhmady:
52 is a valid damage reason, but invalid death reason
explanation here:

https://github.com/openmultiplayer/open.mp/issues/600#issuecomment-1377236916

Hual:
> 52 is a valid damage reason, but invalid death reason explanation here:
> 
> [#600 (comment)](https://github.com/openmultiplayer/open.mp/issues/600#issuecomment-1377236916)

Maybe add a comment in the code about it too

Y-Less:
Oh.  Then yes I agree with documentation.

PR #823, opened 2023-12-31
Sdk files

Y-Less:
Added some missing headers, and split the generic definitions for things like fighting styles and weapon slots out from the "IPlayer" structure file.


SDK/include/game_types.hpp
```diff
@@ -0,0 +1,447 @@
...
+		Player
+	};
+
+	bool spectating;
+	int spectateID;
+	ESpectateType type;
+};
+
+/// The player's name status returned when updating their name
+enum EPlayerNameStatus
```
Hual, line 442: This isn't really a game type

Y-Less: True, a better name is probably required.  They're more just generic structs and enums, but I hate the filename `structs.hpp` or similar, it doesn't tell you anything.

Hual: player_types.hpp or player_structs.hpp?

PR #831, opened 2024-01-04
Add banners & discord configs and send through query

AmyrAhmady:
new configs are now available, example:
```json
{
    "discord": "https://discord.gg/samp",
    "banner_light": "https://github.com/openmultiplayer/graphics/blob/master/rendered/wordmark-light-mono.png?raw=true",
    "banner_dark": "https://github.com/openmultiplayer/graphics/blob/master/rendered/wordmark-coloured.png?raw=true"
}
```
then with query `o`, instead of sending same receiving back, we send these.
max. banner url length is 120 chars.
max. discord url length is 50 chars.

### DIFF
--- a/Server/Components/LegacyNetwork/Query/query.cpp
+++ b/Server/Components/LegacyNetwork/Query/query.cpp
@@ -12,6 +12,8 @@
 const size_t MAX_ACCEPTABLE_HOSTNAME_SIZE = 63;
 const size_t MAX_ACCEPTABLE_LANGUAGE_SIZE = 39;
 const size_t MAX_ACCEPTABLE_GMTEXT_SIZE = 39;
+const size_t MAX_ACCEPTABLE_DISCORD_LINK_SIZE = 50;
+const size_t MAX_ACCEPTABLE_BANNER_URL_SIZE = 120;
 
 template <typename T>
 void writeToBuffer(char* output, size_t& offset, T value)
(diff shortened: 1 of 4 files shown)


edgyaf:
discord connector already uses the `discord` section in `config.json`. would this by any chance cause a conflict?
```json
"discord": {
    "bot_token": "MYBOTTOKEN"
}
```

Y-Less:
So obviously Discord is the chat program du-jour, but there are others, notably Telegram.  It would be good if this was more generic, so you could specify some meta-data for any program - name, link, graphics, etc.  Even an array of them.  Rather than hard-coding to just one.

AmyrAhmady:
This is only to make it appear on launcher like this in bottombar if available
image

We do not have any design or space to put different social links there. Since this is for a gaming community, most of them already use discord, a gamers targeting platform, and it's the first go-to option for gaming communities. For any other social links players can just go to their websites (like facebook? why would we show a facebook link?), but for discord for players (gamers) it's now really an easy way to just access their discord.

Also we want to keep the packet as small as possible because query requests are not authenticated and tied to a player and anyone from anywhere can send a request. having big packet data can be a problem.

Server/Components/LegacyNetwork/Query/query.cpp
```diff
@@ -122,6 +124,38 @@ void Query::buildServerInfoBuffer()
...
 }
 
+void Query::buildExtraServerInfoBuffer()
+{
+	if (core == nullptr)
+	{
+		return;
+	}
+
+	uint32_t discordLinkLength = std::min(discordLink.length(), MAX_ACCEPTABLE_DISCORD_LINK_SIZE);
```
Hual: There's no point in sending an incomplete URL, better to just skip them

Hual (dismissed):

Hual:
> discord connector already uses the `discord` section in `config.json`. would this by any chance cause a conflict?
> 
> ```json
> "discord": {
>     "bot_token": "MYBOTTOKEN"
> }
> ```

As long as the leaf properties are different it wouldn't

Y-Less:
> We do not have any design or space to put different social links there. Since this is for a gaming community, most of them already use discord, a gamers targeting platform, and it's the first go-to option for gaming communities. For any other social 

That's why I mentioned telegram.  It and VK are still very popular alternatives in Russia and other places.

AmyrAhmady:
> That's why I mentioned telegram. It and VK are still very popular alternatives in Russia and other places.

I'll do it in another PR, merging this one as of now

AmyrAhmady:
Current format for configs are:
```json
{
    "discord": {
        "invite": "https://discord.gg/samp"
    },
    "banners": {
        "light": "https://github.com/openmultiplayer/graphics/blob/master/rendered/wordmark-inverse.png?raw=true",
        "dark": "https://github.com/openmultiplayer/graphics/blob/master/rendered/wordmark-inverse.png?raw=true"
    }
}
```

Issue #829, opened 2024-01-03
For some reason GetPlayerGravity gives a value of 0.0 (before it is changed via SetPlayerGravity).


shierru:
# Description
When using the **GetPlayerGravity** function to save the player's gravity before changing it via **SetPlayerGravity**, I found that **GetPlayerGravity** gives `0.0`.

I thought this behavior is not logical, because there is a global gravity value, which is specified in **config.json** (`game.gravity`). And I expected **GetPlayerGravity** to output it (since **SetPlayerGravity** was not used before). Then I opened the source code and noticed that it seems that the value specified in **config.json** should be set by default (correct me if this is wrong).

If my assumption that the **GetPlayerGravity** function should output a value from **config.json** (only if the gravity for the player has not been changed before) is correct, then this is a bug. If not, then I think it would be more logical to have this behavior (again, correct me if I'm wrong).

shierru:
Now we can close this up.

PR #832, opened 2024-01-06
Change gravity type to float

vedran77:
#829

### DIFF
--- a/Server/Source/player_impl.hpp
+++ b/Server/Source/player_impl.hpp
@@ -111,7 +111,7 @@ struct Player final : public IPlayer, public PoolIDProvider, public NoCopy
 	bool toSpawn_;
 	TimePoint lastGameTimeUpdate_;
 	PlayerSpectateData spectateData_;
-	int gravity_;
+	float gravity_;
 	bool ghostMode_;
 	int defaultObjectsRemoved_;
 	bool allowWeapons_;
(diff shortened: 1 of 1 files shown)


shierru:
The fact that the data type is wrong is correct, but the fact that you fixed it with a number (0.008) is wrong. I already mentioned in the Issue that there is a parameter `game.gravity` in **config.json** and it should be the default value.

I think the main problem lies in the data type, which was set incorrectly, because of which the `GetPlayerGravity()` function returned an incorrect value .

vedran77:
Thank you for your feedback. Regarding the instances where I used 0.008f in the code, please note that these values are not intended for actual use; rather, they are included solely for resetting player variables.

The default value (specified in config.json) will be applied each time a player is initialized. You can find the relevant implementation here: [link to the code](https://github.com/openmultiplayer/open.mp/blob/master/Server/Source/core_impl.hpp#L1164).

shierru:
In my opinion it is better to use `0.0` than `0.008`, because then `0.008` becomes a "magic number" and it requires an additional declaration of it as a constant.

vedran77:
You are probably right; the only reason I set it to 0.008 is because I saw in the documentation that the default value is 0.008. However, upon further consideration, it makes sense for it to be 0.0.

Y-Less:
@shierru is correct about using the config value.  As it stands now calling `GetPlayerGravity` on a new player will return the wrong value, where that should really be possible.

vedran77:
I don't know how to reproduce that issue. Here are the things I did:

- Created two commands:
```
YCMD:getgravity(playerid, params[], help) {
    printf("playerid %d - gravity %f", playerid, GetPlayerGravity(playerid));
    return 1;
}

YCMD:setgravity(playerid, params[], help) {
    SetPlayerGravity(playerid, 0.044);
    return 1;
}
```
- Set gravity to 0.01 in my config.json.
- When the player with ID 0 initially connected, they received gravity set to 0.01.
- Then, ID 0 executed the command /setgravity and got gravity set to 0.044.
- After that, ID 1 joined the server and executed the command /getgravity, but the result was 0.01.

What did I miss?

Hual (approved):

Hual:
You didn't miss anything - the gravity is already set to the config one in onPeerConnect()

PR #893, opened 2024-03-12
Fix vehicle teleport bug and revert issue 735

dimmyi:
**1.** Forcing someone to exit spectator mode when destroying a vehicle breaks many systems.

Most admin spectator systems when a player enters a vehicle, the admin starts spectating the player's vehicle, and if some other system, for example storing the vehicle in the garage using DestroyVehicle, would force the admin to stop the mode spectator before even knowing that the player left the vehicle.

**2.** Stream the vehicle driver immediately to avoid the vehicle teleport bug after spectating a vehicle and have a driver. This causes the vehicle to stay in position for a moment and when the driver is StreamIn the vehicle teleports to its position.


NexiusTailer:
Does the first change now make the behaviour like in samp server?

ksenonadv:
Attempting to stream the vehicle right away is indeed the right thing to do.
I'm not sure about 2nd commit, we should keep player in spectating state (client will spectate nothing basically so camera will be around 0 0 0), and reset server side spectating vehicle data.

PR #896, opened 2024-03-16
Boundary between objects on the client (tested)

Locky0:
I ran several tests, completely removing the limitation and using 037, and making this change that worked well on both the 037 and 03DL.
I also did tests on vehicles, which I was unsuccessful at, but I preferred to keep the objects.

### DIFF
--- a/SDK/include/values.hpp
+++ b/SDK/include/values.hpp
@@ -4,7 +4,7 @@
 
 constexpr int MAX_SEATS = 4;
 constexpr int PLAYER_POOL_SIZE = 1000;
-constexpr int VEHICLE_POOL_SIZE = 3000;
+constexpr int VEHICLE_POOL_SIZE = 2000;
 constexpr int CLASS_POOL_SIZE = 320;
 constexpr int OBJECT_POOL_SIZE = 2000;
 constexpr int OBJECT_POOL_SIZE_037 = 1000;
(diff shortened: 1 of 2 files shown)


NexiusTailer:
So you've got the object limit (2k) from 0.3.DL playing on 0.3.7 client version, right? This may be just because you tested it on R3-R5, while on R1-R2 it won't work and the limit will be 1k like 0.3.7 spec says.

ksenonadv:
This obvs won't work at all, server will attempt to use slots > 1000; client won't be able to create those objects.

Locky0:
I carried out tests, in 037 it creates 999 objects.  in 03DL there is more than that.

Locky0:
but using a streamer that directly uses the playerobject, with more than 60k objects.  If the condition that I was able to perform is not there, client 037 would crash when passing the object limit.

Locky0:
> Então você tem o limite de objetos (2k) do 0.3.DL reproduzido na versão do cliente 0.3.7, certo? Isso pode ser apenas porque você testou no R3-R5, enquanto no R1-R2 não funcionará e o limite será de 1k, como diz a especificação 0.3.7.

limit of all versions 0.3.7 is 1000 objects.

AmyrAhmady:
This PR is not valid
Only 0.3DL allows 2k objects. If we remove 1k limit, it wouldn't work for 1k extra objects in 0.3.7, players won't see extra objects and may crash, while server thinks it's good to create from 1000 to 2000, which is problematic, basically means dropping support

Closing it.

Issue #738, opened 2023-10-12
[BEHAVIOUR DIFFERENCE] Client don't get any messages when you enter an invalid nickname


ReshiramZekrom1:
**Describe the bug**
On open.mp you don't get any messages telling you that the chosen nickname isn't valid. This happens either if you use unacceptable characters or if you enter the same nickname of an already online player. You only get the "Server closed connection" message.

**To Reproduce**
1. Choose a nickname with an unacceptable character or enter the same nickname of another already online player
2. Connect to the server

**Expected behavior**
This is what it should happen.

> {FFFFFF}SA-MP {B9C9BF}0.3.7-R5 {FFFFFF}Started
> 
> Connecting to 46.105.43.212:8888...
> 
> Connected. Joining the game...
> 
> CONNECTION REJECTED: Unacceptable NickName
> 
> Please choose another nick between and 3-20 characters
> 
> Please use only a-z, A-Z, 0-9
> 
> Use /quit to exit or press ESC and select Quit Game

**Screenshots and/or logs**
SA:MP behaviour:

<img src="https://i.ibb.co/58zWLD6/sa-mp-005.png" alt="sa-mp-005" border="0">

open.mp behaviour:

<img src="https://i.ibb.co/k2HLSSH/sa-mp-006.png" alt="sa-mp-006" border="0">

**Commit hash in master**
(https://github.com/openmultiplayer/open.mp/commit/441a63a74712003d8a3c05745cc30c93f07922ff)

NexiusTailer:
The same abt this issue, could it be closed if it's solved?

ReshiramZekrom1:
Still present on https://github.com/openmultiplayer/open.mp/actions/runs/8556095715 (https://github.com/openmultiplayer/open.mp/commit/6941fee046cb4c4501236e048d4f2fd0ee0a283c)

PR #900, opened 2024-03-23
fix return 130 rpc, if player have a bad name

atomlin-git:
fix issue #738


Hual (dismissed):

Hual:
While this change makes sense I don't see how it would actually fix the linked issue

atomlin-git:
RPC simply does not have time to reach the player

With out changes: 
image

With changes:
image

Hual:
> RPC simply does not have time to reach the player
> 
> With out changes: image
> 
> With changes: image

But isn't the player being kicked right after that in another function?

atomlin-git:
More better solution in #906

Issue #611, opened 2023-01-11
More sanity checks for config options


NexiusTailer:
Some sanity checks could be added to warn about one of the examples:
* `maxnpc` variable exceeds the maximum number of player slots
* `query` is disabled but `announce` is enabled, so the server will not be visible in masterlist
* More checks on negative values for the most of settings which cannot apply such numbers

Other checks which also can be applied from the server start:
* scriptfiles folder cannot be created for some reason (some scripts rely on this and may work incorrectly)

vsmxd:
Added PR: https://github.com/openmultiplayer/open.mp/pull/1097

PR #901, opened 2024-03-24
add more sanity checks config options

atomlin-git:
Feature #611 

Made:
- print warning, if:
   - 'enable_query' disabled, and 'announce' enabled
   - 'max_bots' >= 'max_players'

- print error, if cant create scriptfiles folder
- fix negative values variables (player_timeout, minimum_connection_time, messages_limit, message_hole_limit, acks_limit, limits_ban_time, weather, time)

### DIFF
--- a/Server/Components/LegacyNetwork/legacy_network_impl.cpp
+++ b/Server/Components/LegacyNetwork/legacy_network_impl.cpp
@@ -876,7 +876,7 @@ void RakNetLegacyNetwork::start()
 	}
 
 	if(*config.getInt("max_bots") >= maxPlayers)
-		core->logLn(LogLevel::Warning, "The 'max_bots' option more or equal how 'max_players' option. Server work can be unpredictable");
+		core->logLn(LogLevel::Warning, "The 'max_bots' option more or equal than 'max_players' option. Server work can be unpredictable");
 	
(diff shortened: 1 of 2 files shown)


NexiusTailer:
Thanks for implementing most of the suggested stuff! But I'd change just a couple of strings which is printed now, they are:
1. `The 'max_bots' option more or equal how 'max_players' option`: "how" -> "than"
2. `Cant create 'scriptfiles' folder`: "Cant create" -> "Cannot access"

Hual:
I think there should also be warnings for the negative options

atomlin-git:
> I think there should also be warnings for the negative options

my solution looks unattractive, tell me how to improve it?

image

Hual:
> > I think there should also be warnings for the negative options
> 
> my solution looks unattractive, tell me how to improve it?
> 
> image

It's fine, no reason to redo it if it's done already

NexiusTailer:
I also thought about this warning message:
`The 'max_bots' option more or equal than 'max_players' option. Server work can be unpredictable`
Basically, the behaviour is predictable, players just can't join in this case. Maybe worth noticing only about "bots option more or equal than players" as the consequences are anyway pretty obvious?

Issue #867, opened 2024-02-14
response parameter in OnDialogResponse


NexiusTailer:
Currently, cheaters may modify dialog response with any integer value from 0 to 255:

![image](https://github.com/openmultiplayer/open.mp/assets/13169094/17e1ac0f-4654-4232-9b99-de7e55b00197)

and it will pass untouched to the pawn script.

I suggest always converting `response` parameter into boolean when processing any dialog response.

Issue #866, opened 2024-02-14
Improve listitem validation


NexiusTailer:
Now listitems in dialogs are validated this way:

https://github.com/openmultiplayer/open.mp/blob/62a2fa95b36b73640a5b4bb330a1663d6d50adc9/Server/Components/Dialogs/dialog.cpp#L158-L162

but it should also consider any other dialog styles (`DIALOG_STYLE_MSGBOX`, `DIALOG_STYLE_INPUT` and `DIALOG_STYLE_PASSWORD`) where listitem always -1 only, and any other values are inacceptable.

The ideal way of validating listitems would be remembering how much items the shown dialog actually has for a player (counting it when ShowPlayerDialog was called, [as an example](https://github.com/Lordzy/safeDialogs/blob/master/pawno/safeDialogs.inc#L141-L160)) and clearly check if the listitem is in the range of 0..[max listitem value] for dialogs with any tablist style; Otherwise check if the listitem is not -1 for any other dialog styles.

PR #902, opened 2024-03-26
improved dialog response checking

atomlin-git:
Maded:

- now, when processing a response to a dialogs next types: "DIALOG_STYLE_LIST", "DIALOG_STYLE_TABLIST_HEADERS", "DIALOG_STYLE_TABLIST" the number of lines in the body of the dialogue is counted, and if the "listitem" value that the player sent is more than we counted - such a response is ignored.

- limited the response parametr value from 0 to 1

Feat #866 and #867

P.S: Implemented through a loop, not a "Impl::count", since in this case the loop should be faster

### DIFF
--- a/Server/Components/Dialogs/dialog.cpp
+++ b/Server/Components/Dialogs/dialog.cpp
@@ -144,11 +144,16 @@ private:
 		{
 			NetCode::RPC::OnPlayerDialogResponse sendDialogResponse;
 			if (!sendDialogResponse.read(bs))
+			{
 				return false;
+			}
 
 			// If the dialog id doesn't match what the server is expecting, ignore it
 			PlayerDialogData* data = queryExtension<PlayerDialogData>(peer);
 			if (!data || data->getActiveID() == INVALID_DIALOG_ID || data->getActiveID() != sendDialogResponse.ID)
+				return false;
+
+			if(sendDialogResponse.Response < 0 || sendDialogResponse.Response > 1)
 				return false;
 
 			if((data->style_ == DialogStyle_PASSWORD || data->style_ == DialogStyle_INPUT || data->style_ == DialogStyle_MSGBOX) && sendDialogResponse.ListItem != -1)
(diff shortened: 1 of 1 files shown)


Server/Components/Dialogs/dialog.cpp
```diff
@@ -167,7 +175,7 @@ class DialogsComponent final : public IDialogsComponent, public PlayerConnectEve
 				&PlayerDialogEventHandler::onDialogResponse,
 				peer,
 				sendDialogResponse.ID,
-				static_cast<DialogResponse>(sendDialogResponse.Response),
+				static_cast<DialogResponse>(sendDialogResponse.Response % 2),
```
Hual: Wouldn't it be better to skip dialog handling for dialogs with Response not 0 or 1?

Server/Components/Dialogs/dialog.cpp
```diff
@@ -144,21 +144,29 @@ class DialogsComponent final : public IDialogsComponent, public PlayerConnectEve
...
-			{
+			if((data->style_ == DialogStyle_PASSWORD || data->style_ == DialogStyle_INPUT || data->style_ == DialogStyle_MSGBOX) && sendDialogResponse.ListItem != -1)
 				return false;
+
+			if((data->style_ == DialogStyle_LIST || data->style_ == DialogStyle_TABLIST || data->style_ == DialogStyle_TABLIST_HEADERS) && data->body_.length() > 0)
+			{
+				unsigned int lines = 0;
+
+				for(unsigned int i = 0 ; i < data->body_.length(); i++)
+					if(data->body_[i] == '\n' && data->body_[i + 1] != '\0')
```
Hual: I don't see the point of the check `data->body_[i + 1] != '\0'` - you can just loop through data->body_.length()-1 and skip the last character

Server/Components/Dialogs/dialog.cpp
```diff
@@ -144,21 +144,29 @@ class DialogsComponent final : public IDialogsComponent, public PlayerConnectEve
 		{
 			NetCode::RPC::OnPlayerDialogResponse sendDialogResponse;
 			if (!sendDialogResponse.read(bs))
-			{
```
Hual: There's no point in this change

Hual (changes requested):

Hual (approved):

ksenonadv (approved):

PR #906, opened 2024-03-31
Fixed: several connection logic issues

uifdev:
1. Clients weren't kicked if the client version doesn't match the server version.
2. Clients were flagged as logged in to the game regardless of the result of OnPeerConnect, meaning they wouldn't be dropped in the event of a failure.

### DIFF
--- a/Server/Components/LegacyNetwork/legacy_network_impl.cpp
+++ b/Server/Components/LegacyNetwork/legacy_network_impl.cpp
@@ -360,11 +360,7 @@ IPlayer* RakNetLegacyNetwork::OnPeerConnect(RakNet::RPCParameters* rpcParams, bo
 			RakNet::BitStream bss;
 			bss.Write(uint8_t(newConnectionResult.first));
 			rakNetServer.RPC(130, &bss, RakNet::HIGH_PRIORITY, RakNet::UNRELIABLE, 0, rid, false, false, RakNet::UNASSIGNED_NETWORK_ID, nullptr);
-
-			if (newConnectionResult.first != NewConnectionResult_VersionMismatch)
-			{
-				rakNetServer.Kick(rid);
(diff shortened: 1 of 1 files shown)


Server/Components/LegacyNetwork/legacy_network_impl.cpp
```diff
@@ -450,19 +445,11 @@ void RakNetLegacyNetwork::OnPlayerConnect(RakNet::RPCParameters* rpcParams, void
...
 			}
 		}
-		else
-		{
-			network->rakNetServer.Kick(rpcParams->sender);
-		}
-		return;
 	}
-
-	network->rakNetServer.Kick(rpcParams->sender);
```
Hual, line 465: Why were the kicks here removed?

atomlin-git:
I don't think these changes make any sense. A kick for an incorrect version will be executed one way or another due to the return of nullptr. Regarding the fact that clients were marked as entering the game, this does not pose any problem. One way or another, if the player does not take the necessary actions, he will be kicked lower.

uifdev:
> A kick for an incorrect version will be executed one way or another due to the return of nullptr.

No, it won't. This PR fixes precisely that bug. Don't comment if you can't read or understand the code.

@Hual Re-added kick calls as per your suggestion, so invalid clients get kicked ASAP.

AmyrAhmady:
Syncing raknet here, since it's related. changes are from this PR: https://github.com/openmultiplayer/RakNet/pull/22

Hual (dismissed):

AmyrAhmady (dismissed):

ksenonadv (dismissed):

AmyrAhmady (approved):

ksenonadv (approved):

PR #907, opened 2024-04-01
Fix FS callbacks

Hual:
As every team member is well aware, open.mp is based on fixes.inc, as established in the project proposal they read and agreed to before joining the team.  Unfortunately a recent commit removed one of these long established features, removing compatibility and breaking scripts in the process.  This is obviously unacceptable, and this PR fixes said mistake.


shr0x:
As every team member is well aware, open.mp is based on fixes.inc, as established in the project proposal they read and agreed to before joining the team. Unfortunately a recent commit removed one of these long established features, removing compatibility and breaking scripts in the process. This is obviously unacceptable, and this PR fixes said mistake.

Hual:
April fools

Issue #882, opened 2024-03-04
messageslimit bug


0724-Tsu:
If server closed connection for player with reason 'messageslimit' in log.txt - player-'phantom' stay on server, but physical player is banned. Player-phantom have ping, skin, all information about him and occupies a slot.

To Reproduce this - reach the 'messageslimit' limit.

akosd127:
#801

AmyrAhmady:
should work in latest nightly build now, closing

https://github.com/openmultiplayer/open.mp/actions/runs/8556095715#artifacts

Issue #801, opened 2023-12-06
Kick bug, maybe duplicate


PazzOnee:
**Describe the bug**
When user get kicked by messageslimit, player start on server online but he is not in game

[2023.12.06 - 18:58:32] [Info] [join] user_name has joined the server (0:ip_address)
[2023.12.06 - 18:59:02] [Warning] client exceeded 'messageslimit' 31.176.176.56:51608 (514) Limit: 500/sec

**To Reproduce**
Spam messages with cheats, idk which exactly

**Commit hash in master**
Release 2 and 3 and every nighly

akosd127:
"When the player exceeds the limit, the following events occur in the game. (see Image 1)
A warning message appears on the console. (see Image 2)
On the server, the player remains active. (see Image 3) The client is no longer needed and can be closed, creating a 'shadow' player.
To remove this lingering player, the ban must be lifted using the 'unbanip' command, followed by reconnecting to the server where an infinite connection attempt will occur.
If the connection attempt fails and the client disconnects, the 'shadow' player is kicked from the server due to a timeout. (see Image 4)"
![1](https://github.com/openmultiplayer/open.mp/assets/15962488/da72f5ac-b280-4b6d-b92a-2fd5b846df96)
![2](https://github.com/openmultiplayer/open.mp/assets/15962488/8c4a01d6-24b3-4eda-a142-99875a1e9ae2)
![3](https://github.com/openmultiplayer/open.mp/assets/15962488/cbc56449-6dbf-45aa-83ba-05197284f31b)
![4](https://github.com/openmultiplayer/open.mp/assets/15962488/1a29fa44-893f-4734-83b6-6bd845e702d4)

PazzOnee:
Fixed in https://github.com/openmultiplayer/open.mp/actions/runs/8556095715, closed

PR #908, opened 2024-04-01
Handle internal immediate forced disconnects

AmyrAhmady:
This PR is basically doing what https://github.com/openmultiplayer/RakNet/pull/21 does but properly by handling forced closed connections/disconnects

Quoted explanation from the said PR:
> If a player exceeds a network limit (e.g. messageholelimit), the player's IP is banned and the connection is immediately closed by RakNet internally. The open.mp server is by design not notified of connections closed internally by RakNet. This causes three issues. First, the player concerned will remain connected in open.mp as a zombie player. Second, once all player IDs are taken, new players are assigned player ID -1. Third, due to missing array boundary checks in the open.mp server at several places, the server will crash when trying to access an array at index -1. This commit fixes all issues by banning the IP only, which results in a connection time out, of which the open.mp server is notified, which deletes the player accordingly.

### DIFF
--- a/Server/Components/LegacyNetwork/legacy_network_impl.cpp
+++ b/Server/Components/LegacyNetwork/legacy_network_impl.cpp
@@ -529,6 +529,12 @@ void RakNetLegacyNetwork::OnRakNetDisconnect(RakNet::PlayerIndex rid, PeerDiscon
 	networkEventDispatcher.dispatch(&NetworkEventHandler::onPeerDisconnect, *player, reason);
 }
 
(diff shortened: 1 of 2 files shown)


AmyrAhmady:
Fixes #882 and #801

ksenonadv (approved):

Hual:
Any idea what the performance impact of attaching a plugin is? Did you test if the problem is fixed?

AmyrAhmady:
Closing this, since we found a better solution, available in here: https://github.com/openmultiplayer/RakNet/pull/22

PR #920, opened 2024-04-22
Recursively load components

myudev:
Easily done QoL change for people who want to structure their components, if there's a config switch needed for whatever reason or any other reason to not do this comment kty <3


Y-Less:
Have you tested how this works with the `include`/`exclude`  fields?  Would doing `"exclude": [ "subdir/component" ]` work?  What about `"exclude": [ "subdir\component" ]` or even `"exclude": [ "subdir" ]`?  Same for `"include"`.

AmyrAhmady:
There's no include, and I believe we should rename "exclude" as well because it's not self explanatory. there is "components" to explicitly load specified components only and nothing else

myudev:
> Have you tested how this works with the `include`/`exclude` fields? Would doing `"exclude": [ "subdir/component" ]` work? What about `"exclude": [ "subdir\component" ]` or even `"exclude": [ "subdir" ]`? Same for `"include"`.

I kept original behaviour to not break people servers suddenly as the exclude option only checks the filename not extension or path and I couldn't find any documentation. However if wanted I could add an option to detect folder paths in exclude, so u can also exclude whole subfolders

Y-Less:
> There's no include, and I believe we should rename "exclude" as well because it's not self explanatory. there is "components" to explicitly load specified components only and nothing else

"components" is the one I meant.

Hual:
```diff
diff --git a/Server/Source/core_impl.hpp b/Server/Source/core_impl.hpp
index be8b3671..692c4886 100644
--- a/Server/Source/core_impl.hpp
+++ b/Server/Source/core_impl.hpp
@@ -1091,13 +1091,18 @@ private:
                                {
                                        if (excludeCfg && !excludeCfg->empty())
                                        {
-                                               p.replace_extension("");
+                                               ghc::filesystem::path rel = ghc::filesystem::relative(p, path);
+                                               rel.replace_extension();
                                                // Is this in the "don't load" list?
-                                               if (std::find(excludeCfg->begin(), excludeCfg->end(), p.filename().string()) != excludeCfg->end())
+                                               const auto isExcluded = [rel = std::move(rel)](const String& exclude)
+                                               {
+                                                       return ghc::filesystem::path(exclude) == rel;
+                                               };
+                                               if (std::find_if(excludeCfg->begin(), excludeCfg->end(), isExcluded)
+                                                       != excludeCfg->end())
                                                {
                                                        continue;
                                                }
-                                               p.replace_extension(LIBRARY_EXT);
                                        }

                                        IComponent* component = loadComponent(p);

```

Here's a diff that allows sub-directories for exclude

Hual (approved):

Issue #912, opened 2024-04-12
OnPlayerSelectedMenuRow, add checks for impossible row values?


BeckzyBoi:
**Describe the bug**
There was an instance reported in my crash log where row from OnPlayerSelectedMenuRow was 157, when maximum number of rows is actually 12 (ranging from 0-11?).

**To Reproduce**
I'm unable to reproduce this, maybe the cause was a fake rpc packet sent by a player?

**Expected behavior**
row should never be anything less than 0 or more than 11.

**Commit hash in master**
Put the commit hash or link to the Github Action of the broken build here.
https://github.com/openmultiplayer/open.mp/actions/runs/8556095715

As far as i'm aware it can happen on released version too, but i am using latest nightly (link above).

PR #922, opened 2024-04-29
added check for possible menu row values

atomlin-git:
issue #912

### DIFF
--- a/Server/Components/Menus/menu.cpp
+++ b/Server/Components/Menus/menu.cpp
@@ -48,6 +48,9 @@ private:
 					data->setMenuID(INVALID_MENU_ID);
 					return false;
 				}
+				
+				Menu* menuData = self.storage.get(data->getMenuID());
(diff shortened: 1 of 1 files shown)


Server/Components/Menus/menu.cpp
```diff
@@ -48,6 +48,9 @@ class MenusComponent final : public IMenusComponent, public MenuEventHandler, pu
 					data->setMenuID(INVALID_MENU_ID);
 					return false;
 				}
+				
+				Menu* menuData = self.storage.get(data->getMenuID());
```
Y-Less: This could be moved above the previous `if`, which would avoid calling the same code twice by using the `Menu` pointer.

We also prefer using braces to single-line ifs.

AmyrAhmady (approved):

AmyrAhmady (approved):

ksenonadv (approved):

PR #923, opened 2024-05-08
RakNet InternalPacketPool & RakPeer::Send improvements and fixes

AmyrAhmady:
1. This PR adds a new implementation of `MemoryPool` from https://github.com/facebookarchive/RakNet/blob/master/Source/DS_MemoryPool.h , to be only and ONLY used for `InternalPacketPool`, by replacing the old `InternalPacketPool` class implementation with new `MemoryPool`. This is because freeing and allocation new InternalPacket on memory wasn't handled properly in old RakNet, specially freeing them, resulting to many issues and crashes randomly in random occasions that we can't even figure out why.
After going through samp-server disassembled code, I realized there are differences, and it's way different than just a simple `free` call or `delete` operator and it looks more like what RakNet v4 offers, so I made it use that. This can potentially fix the memory leak issue we've had since the beginning of open.mp development, that could've been only noticeable on very long runs.

2. Second change is adding new Send functions (also RPC functions, but they're not used, because they don't matter in this case) to take an instance of `NetworkBitStream` for data, instead of passing raw data and number of **bytes**. RakNet has other Send functions that can take `RakNet::BitStream` and this is done because all `Send` functions call `RakPeer::SendBuffered` which take number of **bits** in your data. So using normal `RakPeer::Send` function which take number of **bytes** and not **bits**, meaning it is always aligned and sends extra bits, while it shouldn't really matter on performance and bandwidth level, many servers and protection layers (like custom firewall rules) depend entirely on **bit** size of the data, and can fail when it's aligned to be valid number of **bytes**

### DIFF
--- a/Server/Components/LegacyNetwork/legacy_network_impl.hpp
+++ b/Server/Components/LegacyNetwork/legacy_network_impl.hpp
@@ -131,11 +131,11 @@ public:
 				const PeerNetworkData::NetworkID& nid = netData.networkID;
 				const RakNet::PlayerID rid { unsigned(nid.address.v4), nid.port };
 
-				return rakNetServer.Send((const char*)bs.GetData(), bs.GetNumberOfUnreadBits(), RakNet::HIGH_PRIORITY, reliability, channel, rid, true);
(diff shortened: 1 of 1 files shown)


AmyrAhmady:
closing this for now
not important
but keeping the branch for future reference

PR #998, opened 2024-10-07
add linux build instructions to README.md

aszarth:
Added instructions on how to build on Linux.  
Also included information about sub module folders for users who don't use the `--recursive` option, so they can manually check and download them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,16 @@ cmake -DCMAKE_BUILD_TYPE=Release ..
 make
 ```
 
+## Building on Linux (with Docker)
 
+Replace `/home/user/open.mp` with your actual path in the last command `docker run`
 
+```bash
+cd open.mp
+mkdir build
+cd docker/build_ubuntu-18.04
+docker build -t openmp .
+docker run --rm -ti -v /home/user/open.mp:/omp -w /omp openmp
+```
 
+After that, in `./open.mp/build/Output/Release/Server`, you will find the `omp-server` file and the `components/` folder.


README.md
```diff
@@ -46,6 +46,10 @@ git clone --recursive git@github.com:openmultiplayer/open.mp
 
 Note the use of the `--recursive` argument, because this repository contains submodules.
 
+If you don't use this parameter, you will need to download the files manually and place them in their respective folders, or you will encounter errors during the build process.
```
AmyrAhmady: this **git** repository contains **git** submodules
if they don't know how to use git then it's not really our problem, we shouldn't even mention this *method* of retrieving submodules, because then it's not going to be the right revision

AmyrAhmady (changes requested):

README.md
aszarth: > this **git** repository contains **git** submodules if they don't know how to use git then it's not really our problem, we shouldn't even mention this _method_ of retrieving submodules, because then it's not going to be the right revision

It is not about know how to use git, the command above shows how to download with sub modules...
I just wanted to make life easier to people who don't like to use this parameter and want to download manually, but you are right, if dependencies change we will need to change on README.md too, so i'll remove this part...

AmyrAhmady:
Also one more question, why are you encouraging to use docker manually and directly when there's a `build.sh` script file for anyone to use easily, without requiring them to switch between different ubuntu version folders, also able to set some configurations

PR #1005, opened 2024-10-18
Added toggle for timer set interval & pause/resume

NoPressF:


hiwyn:
the pause stuff is like a `OnPlayerPause`?

NoPressF:
> the pause stuff is like a `OnPlayerPause`?

No, timer pause

Zorono:
i personally use https://github.com/aaamouii/Timer-Fix-plugin since it provides more useful features like interval editing and editing handler and **GetTimerRemainingTime** and **GetTimerInterval** but your PR seems good and promising :D

Issue #1009, opened 2024-10-22
Add IsValidWeapon native.


shierru:
Add a new native function that will check the validity of weapon IDs.

```pawn
native bool:IsValidWeapon(WEAPON:weaponid);
```

PR #1011, opened 2024-10-23
Add IsValidWeapon native (#1009)

shierru:
Unfortunately I haven't figured out how to properly make the changes in the SDK.

I hope you can help me :)


NoPressF:
This is an incorrect check for the validity of weapons because there are no weapons that have an ID (19, 20, 21, 47, 48, 52)

NexiusTailer:
> This is an incorrect check for the validity of weapons because there are no weapons that have an ID (19, 20, 21, 47, 48, 52)

It's not weapons, it's damage reasons. Moreover 19-21 aren't valid both as weapons or reasons

PR #1014, opened 2024-10-27
Implement vehicle's callbacks (OnVehicleTrailerAttach & OnVehicleTrailerDetach)

NoPressF:


NexiusTailer:
I'd like to bump this stale PR just to suggest some different name scheme for these two callbacks: OnTrailerAttach and OnTrailerDetach. This way they'll be in line with the current OnTrailerUpdate.

uifdev:
The callbacks aren't triggered upon using the AttachTrailerToVehicle/DetachTrailerFromVehicle functions.

NexiusTailer:
> The callbacks aren't triggered upon using the AttachTrailerToVehicle/DetachTrailerFromVehicle functions.

They should probably trigger automatically only if both vehicles are unoccupied. Because if the tower (cab) is occupied, the driver should sync this event by himself.

PR #1026, opened 2024-12-07
omp user detection & server to client encryption for omp users

AmyrAhmady:
i dont like poles
385371624980873228_512x512

### DIFF
--- a/Server/Components/LegacyNetwork/legacy_network_impl.cpp
+++ b/Server/Components/LegacyNetwork/legacy_network_impl.cpp
@@ -313,7 +313,7 @@ enum LegacyClientVersion
 	LegacyClientVersion_03DL = 4062
 };
 
-IPlayer* RakNetLegacyNetwork::OnPeerConnect(RakNet::RPCParameters* rpcParams, bool isNPC, StringView serial, uint32_t version, StringView versionName, uint32_t challenge, StringView name, bool isUsingOfficialClient)
+IPlayer* RakNetLegacyNetwork::OnPeerConnect(RakNet::RPCParameters* rpcParams, bool isNPC, StringView serial, uint32_t version, StringView versionName, uint32_t challenge, StringView name, bool isUsingOmp, bool isUsingOfficialClient)
 {
 	const RakNet::PlayerID rid = rpcParams->sender;
 
@@ -344,6 +344,7 @@ IPlayer* RakNetLegacyNetwork::OnPeerConnect(RakNet::RPCParameters* rpcParams, bo
 		params.bot = isNPC;
 		params.serial = serial;
 		params.isUsingOfficialClient = isUsingOfficialClient;
+		params.isUsingOmp = isUsingOmp;
 		newConnectionResult = core->getPlayers().requestPlayer(netData, params);
 	}
 	else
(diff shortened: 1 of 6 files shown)


AmyrAhmady:
Some explanations:
For connection request packet SAMP uses 3 bytes (one for packet id, the other two for cookie seed I guess...)
I've changed this for omp launcher users and now it's 14 https://github.com/openmultiplayer/RakNet/compare/7bfb64ad99584ba2efcff68a2e096f6edc276291...20bd25b406a60bc72d1e48610f72e88c94962aae#diff-5ebcaa4fdb0b73ba98c95b61d3428b26f7e2297d05bf2e0287048bad76be2834R4103

Which is handled here: https://github.com/openmultiplayer/RakNet/compare/7bfb64ad99584ba2efcff68a2e096f6edc276291...20bd25b406a60bc72d1e48610f72e88c94962aae#diff-e5d7976b6cdaaee7a6be2e173c0251ae68161545e8023844925bceb5abf20979R677
Basically checks if length is 14, if it is, check if it has `omp` bytes in it, if so, get the data we need out of it, which is omp version and player's encryption randomized seed.

Then it configures player settings and then reply back to client with this https://github.com/openmultiplayer/RakNet/compare/7bfb64ad99584ba2efcff68a2e096f6edc276291...20bd25b406a60bc72d1e48610f72e88c94962aae#diff-e5d7976b6cdaaee7a6be2e173c0251ae68161545e8023844925bceb5abf20979R656
This is client's way of knowing it has connected to an omp server. that "magic" number is only there so there's no conflict between our usage of this packet ID and other people's third party plugins. Client also checks server version and warns the user if there's a mismatch

Encryption is pretty easy to understand as well https://github.com/openmultiplayer/RakNet/compare/7bfb64ad99584ba2efcff68a2e096f6edc276291...20bd25b406a60bc72d1e48610f72e88c94962aae#diff-e5d7976b6cdaaee7a6be2e173c0251ae68161545e8023844925bceb5abf20979R328
It's just the same as client to server encryption, instead I've made it much simpler by only having a XOR against player's random seed value. also have a checksum thingy just like client to server again, it's only there to tell both sides the arriving packet is valid and needs to be processed. if checksum fails it means it's not good for processing.


Now with all that, it means it is required to know whether a user is using omp launcher or not, and since we keep track of that, I also added `IsPlayerUsingOmp` for both SDK users and Pawn users

AmyrAhmady:
Do not merge this yet by the way, I could be changing some stuff first.

NexiusTailer:
Just one little note about the naming of native added. We already have IsPlayerUsingOfficialClient and now we also will have another to determine that a player is playing from omp launcher/client mod. But `IsPlayerUsingOfficialClient` and `IsPlayerUsingOmp` may seem mutually exclusive for a scripter, so I'd suggest a bit different name, like more specific `IsPlayerUsingOmpLauncher` or `HasPlayerOmpClientMod`.

Hual:
Client -> server should be encrypted too ideally

AmyrAhmady:
Update
Now omp uses a new value for petarded (yes) and it's `0x6D70`, bytes for `mp`, in ID_OPEN_CONNECTION_REQUEST packet
then now in a new way, instead of using gpci, it's just a random uint32 encryption seed, and it is generated on the server side then later sent, right after receiving ID_OPEN_CONNECTION_REQUEST 

The reason behind these changes is so many hosting providers have their custom set of firewall rules so I'm not able to send or customize packets from client to server in any way I want, so I had to re-use current packet without customization (like ID_OPEN_CONNECTION_REQUEST)

Also latest raknet changes are synced too so this PR contains race condition fixes too

AmyrAhmady:
force pushed for rebasing

PR #1032, opened 2024-12-13
Add help message when attempting to load a script without arguments

xerox8521:
This PR checks if no arguments were passed when attempting to (re-|un-)load a (filter-)script or gamemode and prints a help message instead of just failing to load the script.


AmyrAhmady (approved):

AmyrAhmady:
formatting is not correct, please check clang format ci logs and fix it

PR #1052, opened 2025-01-26
Fix: Avoid a update unstreamed cars by passenger sync

povargek:
Fix: Avoid a update unstreamed cars by passenger sync

This patch fixes cheat ability to update by passenger sync to unstreamed cars for player. Old SA:MP bug..


povargek:
delete, clang fail

PR #1053, opened 2025-01-26
Fix: Avoid a update unstreamed cars by passenger sync

povargek:
Fix: Avoid a update unstreamed cars by passenger sync

This patch fixes cheat ability to update by passenger sync to unstreamed cars for player. Old SA:MP bug..


NexiusTailer:
There is actually a potential reason why kalcor left passenger sync without a stream check: train carriages always marked as unstreamed in samp so you should exclude them from the main check.

povargek:
I will think how to resolve it..

AmyrAhmady:
Solve the conflicts, there's now a new internal function called IsTrainCarriage, you can check it against that

povargek:
Ok, i check it

NexiusTailer:
This is already fixed in #1113

PR #1066, opened 2025-03-04
Implement object streaming

Lyuu17:


AmyrAhmady:
Hi there,  

While we’re happy and thankful for the time and effort you put into this and contributed to this project, we would still appreciate it if you discussed such changes/additions with the team first, either through our Discord or GitHub issues.  

There are a few reasons why I **can't** merge this pull request.  

1. You need to heavily test this with the streamer plugin. There's no guarantee this implementation works fine. We **need** to support legacy plugins and scripts—not just for backward compatibility, but also due to the lack of an actual unlimited streaming object pool (which the streamer plugin provides).  

2. Judging by the code and changes, it looks like you're following the same style used in the Pickups component. This approach is not only horribly done and suboptimal but also *somewhat* broken. Currently, you're looping through **all** objects for **all** players in every `onPlayerUpdate` call. This means that every time an on-foot sync is received, or a driver, passenger, or spectator sync occurs, you're looping through **all** objects created for everyone—not just the specific ones for that player.  

    Imagine a scenario where each player has reached their object limit (1,000 objects), and you have **1,000** players. For **one** player's `onPlayerUpdate` event call, you're looping through **1,000 × 1,000** objects. Now, imagine all **1,000** players are sending packets with an update rate of 40ms—**200** times per second for the `onPlayerUpdate` event call per player.  

    This means you're iterating **1,000,000 × 200 × 1,000**, which results in **200,000,000,000** iterations per second. This absolutely destroys the server’s performance. Obviously, not all servers have 1,000 players, and not all players have all 1,000 objects streamed (though only a few don’t—the rest do). However, testing it with Pickups in action on a large server shows that this is horribly inefficient and terribly implemented. In fact, the Pickups component is the one that actually needs fixing.  

3. The third reason is that you're not really improving anything here. You're only streaming 1,000 objects, not breaking the limit. You’re introducing player objects in a new way, but we already *have* player objects with exactly the same limitations. So, with all due respect, this is a very, *very* useless change because it doesn’t provide anything new or additional—it only introduces massive performance issues and potential conflicts with the streamer plugin.  


If there's anything you want to add or discuss, you can continue replying to this PR or discuss it on our discord.
https://discord.gg/samp

Lyuu17:
i though the streamer plugin wasn't supported lol i couldn't find the open.mp sdk in there

but yeah you're right this is really bad i didn't want to completely rework everything and i though the pickup one was ok but it seems not

tysm :)

AmyrAhmady:
Streamer (and all the other samp plugins) are supported without a doubt, except those that are doing memory hacking, such as YSF, FCNPC, and etc. We provide backward compatible api so you don't even have to do anything, just as you do with samp, put them in plugins folder and add them to your config file

PR #1088, opened 2025-05-19
Fixed "reverse search for NPC id in player pool"

NoPressF:


Server/Source/player_pool.hpp
```diff
@@ -1695,7 +1695,8 @@ struct PlayerPool final : public IPlayerPool, public NetworkEventHandler, public
 
 		if (params.bot)
 		{
-			for (auto index = storage.Capacity - 1; index >= 0; --index)
+			const auto maxPlayers = core.getConfig().getInt("max_players");
```
AmyrAhmady: make it static, we don't need to retrieve it every single time. it's a pointer to this config variable so even if it changes during runtime, having a static pointer to it doesn't affect the value

AmyrAhmady (changes requested):

Server/Source/player_pool.hpp
NoPressF: Fixed in the last commit

ksenonadv (approved):

AmyrAhmady (approved):

PR #1091, opened 2025-05-21
Pack Vehicle and StreamInVehicle more tightly

PTemuri:
Minor size optimizations
Pack Vehicle and StreamInVehicle more tightly
Size decrease: for Vehicle from 480 bytes to 464 bytes, for StreamInVehicle from 116 bytes to 108 bytes
May be useful for future development, to avoid crossing alignment boundary
No functional changes whatsover

### DIFF
--- a/Server/Components/Vehicles/vehicle.hpp
+++ b/Server/Components/Vehicles/vehicle.hpp
@@ -48,14 +48,15 @@ private:
 	uint8_t landingGear = 1;
 	bool respawning = false;
 	bool detaching = false;
+	bool beenOccupied = false;
 	FlatHashSet<IPlayer*> passengers;
 	HybridString<16> numberPlate = StringView("XYZSR998");
 	uint8_t objective;
 	uint8_t doorsLocked;
+	uint8_t sirenState = 0;
 	VehicleDeathData deathData;
 	TimePoint timeOfSpawn;
 	TimePoint lastOccupiedChange;
-	bool beenOccupied = false;
 	Vector3 velocity = Vector3(0.0f, 0.0f, 0.0f);
 	Vector3 angularVelocity = Vector3(0.0f, 0.0f, 0.0f);
 	TimePoint trailerUpdateTime;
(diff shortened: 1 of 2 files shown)


xgreatxfortunex333:
@AmyrAhmady 
image
What even is this PR

AmyrAhmady:
It's packing the structures so we don't waste memory due to assigned paddings during allocations.
It is also good for data locality and caching as well

xgreatxfortunex333:
> Minor size optimizations
> Pack Vehicle and StreamInVehicle more tightly
> Size decrease: for Vehicle from 480 bytes to 464 bytes, for StreamInVehicle from 116 bytes to 108 bytes
> May be useful for future development, to avoid crossing alignment boundary
> No functional changes whatsover

so saving (at max vehicles) **0.032 megabytes** matters now?

I got the cache part but memory cmon

AmyrAhmady:
It isn't about only those two structure members. All of them are done like and those two weren't following. 
Just because the size is small it doesn't mean we should leave it like and not follow how it has been done literally everywhere for other members, to have proper alignment.

Also I said "during allocation", maybe I wasn't clear enough. more space to allocate (even small) = less impact on performance
Specially with objects that are allocated very frequently.

But the whole idea of this PR was to fix something that is done everywhere and it was missing for those two members. I we didn't care about it and all the other members had byte paddings then the size difference wouldn't have been that small for you to nitpick it

xgreatxfortunex333:
> It isn't about only those two structure members. All of them are done like and those two weren't following. Just because the size is small it doesn't mean we should leave it like and not follow how it has been done literally everywhere for other members, to have proper alignment.
> 
> Also I said "during allocation", maybe I wasn't clear enough. more space to allocate (even small) = less impact on performance Specially with objects that are allocated very frequently.
> 
> But the whole idea of this PR was to fix something that is done everywhere and it was missing for those two members. I we didn't care about it and all the other members had byte paddings then the size difference wouldn't have been that small for you to nitpick it

I just asked, because usually on the official discord server people would burn you alive if you talked about megabytes of memory in your script yet here 32kb was a concern, sorry~~

AmyrAhmady:
It's fine
As I said, it's more like following a rule so we make sure it's the same everywhere in every structure
it's not noticeably for exact changes in this PR, but it's definitely noticeable when every structure tries to follow this rule

PR #1100, opened 2025-07-01
Fixed: train passenger sync

uifdev:
This PR fixes the following issues related to train passenger sync:
- Players inside a train carriage appear paused; OnPlayerUpdate is not triggered and GetPlayerVehicleID returns 0.
- Entering/exiting a train carriage does not trigger OnPlayerEnter/OnPlayerExitVehicle.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,7 +33,7 @@ on:
 jobs:
   build-windows:
     name: Windows build
-    runs-on: windows-2019
+    runs-on: windows-latest
 
     strategy:
       matrix:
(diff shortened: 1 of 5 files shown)


beckzyb:
Is OnPlayerEnter/ExitVehicle called in SA-MP? I wouldn't have imagined they would be with the player being put into the carriage without entering/exiting animations being performed. Regardless of that, you're right that OnPlayerStateChange and OnPlayerUpdate should be.

NexiusTailer:
Yes, OnPlayer(Enter/Exit)Vehicle called for train carriages when player press F key to enter them by himself, just like with any other vehicles.

PR #1113, opened 2025-08-10
One big beautiful pull request

NexiusTailer:
1. Fixes two critical issues in vehicle (driver) sync processing: they're both caused by wrong code order when pure validation checks which drops the packet executed *after* some side effects are applied, but those checks must be *before* them. This way it prevents partial updates from invalid driver sync. The first place was inside PlayerVehicleSyncHandler and the second in updateFromDriverSync.
2. Fixes critical behavior difference when passenger sync could be passed with any passenger position a cheater want to spoof. SA-MP server ignores what was received by the sender and rewrites the new passenger position with the one his vehicle has now. This change prevents areas for abuse in the existing SA-MP scripts: for example, many anticheats don't check passengers for teleport hack because according to original SA-MP behavior the passenger can update his position in vehicle only if it has no driver, via unoccupied sync.
3. Fixes passenger sync spoofing for non-streamed vehicles, considering train carriages ([#1053](https://github.com/openmultiplayer/open.mp/pull/1053)).
4. Drops unoccupied sync if the cheater send vehicle update from 0 seat being actually a passenger (the check for not being a passenger and updating unoccupied vehicle from passenger seats already was presented, though).
5. Normalizes trailer quaternion from the server side since samp clients don't do it specifically for trailer sync for some reason, so it can be abused by cheaters like they do with unoccupied sync roll/direction matrix.

### DIFF
--- a/Server/Components/Vehicles/vehicle.cpp
+++ b/Server/Components/Vehicles/vehicle.cpp
@@ -164,6 +164,13 @@ bool Vehicle::updateFromDriverSync(const VehicleDriverSyncPacket& vehicleSync, I
 		return false;
 	}
 
+	if (vehicleSync.TrailerID && trailer && trailer->getID() != vehicleSync.TrailerID)
+	{
+		// The client instantly jumped from one trailer to another one.  Probably a cheat, so don't
+		// allow it.
+		return false;
+	}
+
 	pos = vehicleSync.Position;
 	rot = vehicleSync.Rotation;
 	velocity = vehicleSync.Velocity;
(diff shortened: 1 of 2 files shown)


RonaldoBueno16 (approved):

Woet (approved):

itsneufox (approved):
Looks good

edgyaf (approved):
✅ Vertified

DEntis-T (approved):

midosvt (approved):
Wow, this PR is like taking a messy plate of spaghetti code, untangling the noodles, and putting them back in order before anyone chokes on them. 🍝

The fixes make total sense -- moving validation checks before side effects is the programming equivalent of washing your hands before eating, not after. Also, I appreciate the cheater smackdown -- no more ‘free teleport to anywhere’ passenger perks. ✈️

Overall: looking really good. Clean, logical, and it closes some pretty nasty loopholes. Thanks for making vehicle sync less of a clown car. 🚗🤡

NotUnlikeTheWaves (approved):
waowieee

Sreyas-Sreelal (approved):

equinnoxy (approved):

thelante (approved):
gud

KW46 (approved):

AmyrAhmady (approved):

PR #1129, opened 2025-10-18
Add onTick callback for PawnEventHandler

Kmarco68:
Right now async plugins need users to setup timers manually? didnt found other sulution.This adds an optional onTick callback so plugins can handle callbacks automatically without requiring script timers.

Depends on SDK PR: openmultiplayer/open.mp-sdk#57.

Changes:
SDK: Added onTick() to PawnEventHandler (default does nothing)
 Component: Call eventDispatcher in main.cpp onTick

### DIFF
--- a/Server/Components/Pawn/main.cpp
+++ b/Server/Components/Pawn/main.cpp
@@ -256,6 +256,13 @@ public:
 	{
 		PawnManager::Get()->pluginManager.ProcessTick();
 		PawnManager::Get()->ProcessTick(elapsed, now);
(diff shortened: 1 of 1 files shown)


AmyrAhmady:
I'm confused. If you are making a component, you already have an onTick from CoreEventHandler
You can just use that in your third party component instead of something that does exactly the same.

And if you are making a legacy samp plugin, you have to use ProcessTick exposed function.

Please elaborate on why this is needed

PR #1138, opened 2025-11-07
Add null check for vehicle creation

temajm:


AmyrAhmady:
not needed anymore due to how this process of creating trains has changed

PR #1139, opened 2025-11-10
Force train to be 4 consecutive ids

AmyrAhmady:


0x617374726F:
I think at least cc6268b should be merged immediately. I'd appreciate that since it fixes my callback return values.

ksenonadv (approved):

PR #1141, opened 2025-11-13
fix players being able to join the server even if the slot limit has been reached for max_players/max_bots

### LINKED ISSUE #1137: Players can join the server even if the slot limit has been reached (max_players/max_bots)
**Describe the bug**
NPCs take slots used by regular players, so the slots count in server query decreases correctly. `max_players` 110 with 10 NPCs connected (`max_bots` set to `10`) displays 100 slots in server query. Despite this, players still can join up to 110, occupying higher slots (so in result it can display things like 110/100 in query).

**To Reproduce**
1. Set `max_players` to 1 and `max_bots` to 1.
2. Run this script:
```pawn
#include <open.mp>

main() {}

public OnGameModeInit()
{
    NPC_Create("test");
}
```
3. Try to connect as regular player - it shouldn't be allowed, but the player can join fine.

**Expected behavior**
Player can't connect. Case above describes situation where there are 0 slots available for regular players, but same thing applies to higher values like 110 slots with 10 NPCs.

**Screenshots and/or logs**
`max_players = 2` , `max_bots = 1` (1 NPC connected, 2 regular players managed to connect):
Image

**Commit hash in master**
https://github.com/openmultiplayer/open.mp/actions/runs/18908843333

itsneufox:
fixes #1137

### DIFF
--- a/Server/Source/player_pool.hpp
+++ b/Server/Source/player_pool.hpp
@@ -48,6 +48,7 @@ struct PlayerPool final : public IPlayerPool, public NetworkEventHandler, public
 	bool* allowInteriorWeapons_;
 	bool* logConnectionMessages_;
 	int* maxBots;
+	int* maxPlayers;
 	StaticArray<bool, 256> allowNickCharacter;
 	TimePoint lastScoresAndPingsCached;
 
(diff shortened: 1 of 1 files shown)


itsneufox:
Closing since there's a better fix as discussed on Discord

PR #1144, opened 2025-11-19
Fix event handling

0x617374726F:


AmyrAhmady:
rebase your branch please

AmyrAhmady (approved):

0x617374726F:
Sure. I'm still having an issue though that might be related to this. Using my omprs component, dialogs won't show up if invoked from within on_player_connect callback. It works though if invoked by a command. Note that it didn't work before that PR either.

AmyrAhmady:
Ask omprs dev to test it, I'll test in my own test environment too

0x617374726F:
On omprs side, I am able to workaround the issue by deferring the show dialog invocation to the next tick. That's what we actually do for quite some API calls that are likely to cause new events coming in. In general, this enables us to comply with the borrow checking rules.
But I think that this problem is about something different and not multiple mutable references being held, since the component won't panic in this case.
Therefore, I suspect that the bug can be reproduced by using capi directly.

AmyrAhmady:
The issue is omprs adds event handlers with highest priority instead of default value. It's not happening when it's set to FairlyLow or Lowest. Yet this can be prevented in another way, which I'm making a PR for. @0x617374726F

AmyrAhmady:
Fixed by #1145

PR #1146, opened 2025-11-21
Fix for random occurrence of player state none

### LINKED ISSUE #1008: Two critical behavior diffs which break ac checks
**Describe the bug**
There are a couple of problems in omp server behavior which produce false detections for the server-side anticheat and actually can be (and should be) fixed on the side of omp server itself.

* [x] The first issue appears when the player was in spectating mode and leaves from it, the server instantly sets him `PLAYER_STATE_NONE` which must not be used anywhere except as a bare "just-after-connection" state (when no any other state were set before), because this is the only documented behavior in samp server for this state. Setting it between two other states while playing leads to ruining anticheat logic if it tries to make sane checks (like checks for skipping `PLAYER_STATE_SPAWNED`, when cheaters send onfoot sync just after connection). The place where this kludge is located is probably [here](https://github.com/openmultiplayer/open.mp/blob/01d70fb3325c9027d640bc263182f33515ee5099/Server/Source/player_impl.hpp#L630).
* [x] The second issue was found recently, but seems it persist for a long time too, it's when the server gives a player any non-bullet weapon (for example, brass knuckles or any melee like chainsaw, flowers etc) with ammo set to 0. After player received such weapon in hands, he can switch his current weapon to any other slot and then switch back to new weapon to hold it again. Just after these actions `GetPlayerWeaponData` and other functions will return 0 as server cleared the whole slot from both weapon and ammo, probably because ammo was already 0 so it considers the weapon also disappeared and thus can be cleared from data, but it's not.

**To reproduce the first issue**
1. Toggle a player spectating mode, set some other player for him to spec on
2. Toggle it off by the server or wait until the one he's spectating on will left the game, so our player will be spawned
3. Log `OnPlayerStateChange`, it usually show that his state was changed to 0 and only then to the real one after it

**To reproduce the second issue**
1. Give player some non-bullet weapon with ammo 0:
```pawn
GivePlayerWeapon(playerid, 9, 0); // Chainsaw
```
2. Let him switch his holding weapon back and forth
3. Obtain values from `GetPlayerWeaponData`

**Release version**
Latest v1.3.1.2748, but both issues also exist on older builds

NexiusTailer:
This should finally fix #1008 with pretty critical behavior difference, when omp server as a workaround sets the player state to `PLAYER_STATE_NONE` when player leaves spectating. As the comment says the reasons:
https://github.com/openmultiplayer/open.mp/blob/59fd6e558732ea8f577d63e988c7372b0dc44424/Server/Source/player_impl.hpp#L635-L638
it is supposed to update the player state to pass further checks inside RequestClass or RequestSpawn events, one of which will be called by player when he exits spectating.

But what we see in those places are only such checks:
https://github.com/openmultiplayer/open.mp/blob/59fd6e558732ea8f577d63e988c7372b0dc44424/Server/Components/Classes/classes_main.cpp#L156-L160
^ in PlayerRequestClassHandler, and:
https://github.com/openmultiplayer/open.mp/blob/59fd6e558732ea8f577d63e988c7372b0dc44424/Server/Source/player_pool.hpp#L64-L68
^ in PlayerRequestSpawnRPCHandler

So, if player still has [PLAYER_STATE_SPECTATING](https://open.mp/ru/docs/scripting/resources/playerstates), it **will** pass both of those checks, since spectating state is not in the range of forbidden states which are blocked to call all the mentioned events.

Meanwhile having manually update player states to random ones leads to behavior differences, and in this case it's pretty critical one since it break many protections against state change spoofing and thus falsely triggers on fair players ([#325](https://github.com/NexiusTailer/Nex-AC/issues/325) as example). Not to say that such actions are also undocumented and doesn't match SA-MP behavior for any other scripts which aren't designed to consider similar server behavior surprises.

### DIFF
--- a/Server/Source/player_impl.hpp
+++ b/Server/Source/player_impl.hpp
@@ -631,12 +631,6 @@ struct Player final : public IPlayer, public PoolIDProvider, public NoCopy
 			toSpawn_ = true;
 			spectateData_.type = PlayerSpectateData::ESpectateType::None;
 			spectateData_.spectateID = INVALID_PLAYER_ID;
-
-			// When client exits spectate mode it will attempt to request a class or to spawn
-			// but the server still thinks the it is in it's previous state.
-			// This can lead to those requests being rejected.
-			// This state update will fix that.
-			setState(PlayerState_None, false);
 		}
 		else
 		{


NexiusTailer:
Tested those changes for a week on a test server with different scenarios, so now removing 'draft' from the PR.

v0UwUK6CmCrjVB1PAtRb:
Looks good to me.

This check was [causing issues on my server](https://github.com/NexiusTailer/Nex-AC/issues/325) when running Nex-AC. Namely, players were being wrongfully kicked from the server if they died whilst in a vehicle, which clearly was far from ideal.

AmyrAhmady (approved):

AmyrAhmady:
I just it doesn't break anything
image

NexiusTailer:
I also was surprised when realized that I don't need to even bother deepseek to implement it somehow differently, since it just wasn't needed at all

PR #1156, opened 2025-12-12
Couple of fixes for NPC component

### LINKED ISSUE #1155: Significant NPC component behavior diffs from FCNPC
While chatting in one of ru community resources, one guy has noticed a few significant behavior differences between omp NPC component and FCNPC plugin when porting his gamemode (which heavily uses NPCs, so it should all be good catches because it covers the most of ported NPC features). The reason that I'm reporting it now is just because I summarized it in a brief list instead of tldr claims, each of the point seems pretty important tho:

1. `NPC_SetFacingAngleToPos` and `NPC_SetFacingAngleToPlayer` (based on [FCNPC_SetAngleToPos](https://github.com/ziggi/FCNPC/blob/master/sampsvr_files/FCNPC.inc#L134) and [FCNPC_SetAngleToPlayer](https://github.com/ziggi/FCNPC/blob/master/sampsvr_files/FCNPC.inc#L135)). Those two made life much more easier by that it automatically calculates the angle between the NPC and other player/position and sets the NPC this calculated angle.

2. Make `NPC_Respawn` to be able respawning NPCs [which are dead](https://github.com/openmultiplayer/open.mp/blob/673fc12fb71faef7ee4c1d78f42f0d11bbe6003f/Server/Components/NPCs/NPC/npc.cpp#L279), since FCNPC [can do this](https://github.com/ziggi/FCNPC/blob/e6a9311e002c174f5f0238e0eae26354182de521/src/CPlayerData.cpp#L200) and pretty much logic was built on this fact. `NPC_Respawn` still has sense to be separated from `NPC_Spawn`, because it [re-applies many things](https://github.com/openmultiplayer/open.mp/blob/master/Server/Components/NPCs/NPC/npc.cpp#L305-L330) to the NPC when you respawn him instead of just spawn. Scripters are aware of this difference and sometimes use Respawn on dead NPCs, so it's an issue now (not to mention that ability to respawn dead NPCs with `FCNPC_Respawn` was also [documented](https://github.com/ziggi/FCNPC/wiki/Natives#fcnpc_respawnnpcid)).

3. Make [default NPC move type](https://github.com/openmultiplayer/omp-stdlib/blob/master/omp_npc.inc#L90-L96) like it was in FCNPC (probably it's better to be `NPC_MOVE_TYPE_AUTO`). Currently there are many movement glitches with it when porting, if using the default parameter value in the most of movement natives (you can see some of the cases [here in this demo](https://streamable.com/4uil43)).

NexiusTailer:
This should fix #1155, more specific its second item in the issue list.
Apart from that, it adds a clear error for invalid playback, which was ported from FCNPC playbacks implementation

### DIFF
--- a/Server/Components/NPCs/NPC/npc.cpp
+++ b/Server/Components/NPCs/NPC/npc.cpp
@@ -1649,7 +1649,7 @@ bool NPC::startPlayback(StringView recordName, bool autoUnload, const Vector3& p
 		stopPlayback();
 	}
 
-	playback_ = new NPCPlayback(recordName, playbackPath_, autoUnload, npcComponent_->getRecordManager());
+	playback_ = new NPCPlayback(recordName, playbackPath_, autoUnload, npcComponent_);
 	if (!playback_ || !playback_->isValid())
 	{
 		if (playback_)
@@ -1681,7 +1681,7 @@ bool NPC::startPlayback(int recordId, bool autoUnload, const Vector3& point, con
 		stopPlayback();
 	}
 
-	playback_ = new NPCPlayback(recordId, autoUnload, npcComponent_->getRecordManager());
+	playback_ = new NPCPlayback(recordId, autoUnload, npcComponent_);
 	if (!playback_ || !playback_->isValid())
 	{
 		if (playback_)
--- a/Server/Components/NPCs/NPC/npc.cpp
+++ b/Server/Components/NPCs/NPC/npc.cpp
@@ -274,9 +274,9 @@ void NPC::spawn()
 
 void NPC::respawn()
 {
-	// Make sure the player is already spawned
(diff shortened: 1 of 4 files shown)


NexiusTailer:
As for the logging error message, I'm not sure how can I fix the current compile error. Any help is appreciated...

Also please guide if it's more correct to make it with %s + c_str() method, like:
```C++
core->logLn(LogLevel::Error, "[NPC] Playback initialization failed. Playback file '%s' not found.", fullPath.c_str());
```
or leave as it's now:
```C++
core->logLn(LogLevel::Error, "[NPC] Playback initialization failed. Playback file '%.*s' not found.", PRINT_VIEW(fullPath));
```

AmyrAhmady (approved):

PR #1157, opened 2025-12-16
Handle spawn with no available classes better

NexiusTailer:
This is going to explicitly handle the case when a player gonna spawn from class selection and no class is available.

### DIFF
--- a/Server/Components/Classes/classes_main.cpp
+++ b/Server/Components/Classes/classes_main.cpp
@@ -181,6 +181,11 @@ private:
 					player_data->default_ = false;
 				}
 			}
+			else if (player_data)
+			{
+				player_data->cls = *used_class;
+				player_data->default_ = true;
+			}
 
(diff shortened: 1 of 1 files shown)


NexiusTailer:
Here's what it fixes in practice:
https://discord.com/channels/231799104731217931/966398440051445790/1449775249145991329

iamkovshevoy:
In my opinion it would be better if they removed the class system or gave the option to disable it. Oh, those bugs with F4 + RequestClass/Spawn

NexiusTailer:
> In my opinion it would be better if they removed the class system or gave the option to disable it.

There are 2 types of issues/proposals: those which can be done on client (not relevant currently) and those which can be changed in the current server itself. This PR fixes a real behavior diff in omp server, so it's the second case.

AmyrAhmady (approved):

Issue #1143, opened 2025-11-16
Bug with Cyrillic alphabet flips every 4 bytes in format() and %q


povargek:
In format() function %q breaks the Cyrillic alphabet, flips every 4 bytes. If you add an ASCII symbol first, for example, a space, the problem is solved.

There is no such problem with %q in standard SA:MP server
```
{ // %q
  new string[100];
  new name[] = "Привет мир";
  format(string, sizeof(string), "Test string '%q'", name);
  printf(string);
}

{ // space + %q
  new string[100];
  new name[] = " Привет мир";
  format(string, sizeof(string), "Test string '%q'", name);
  printf(string);
}

{ // %s
  new string[100];
  new name[] = "Привет мир";
  format(string, sizeof(string), "Test string '%s'", name);
  printf(string);
}
```

OUTPUT
2025/07/25 17:19:31 [Info] Test string '**вирПм те**'
2025/07/25 17:19:31 [Info] Test string ' **Привет мир**'
2025/07/25 17:19:31 [Info] Test string '**Привет мир**'

SimoSbara:
Hello, which file encoding are you using in your source code?

AmyrAhmady:
Fixed and will be available in the next release

PR #1159, opened 2025-12-19
Fix string escape with %q in format()

SimoSbara:
This PR solves #1143. 
I reimplemented the same algorithm as before but without the conversion to a char string: `std::string strArg`.

### DIFF
--- a/Server/Components/Pawn/format.cpp
+++ b/Server/Components/Pawn/format.cpp
@@ -783,19 +783,21 @@ reswitch:
 			}
 			if (argLen > 0)
 			{
-				//using only cell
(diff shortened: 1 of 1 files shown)


AmyrAhmady:
image

You can also run `format.bat` available in main folder (you need to have clang format installed) for auto formatting

SimoSbara:
> image
> You can also run `format.bat` available in main folder (you need to have clang format installed) for auto formatting

Thanks, I will use clang-format tool directly, I'm on Debian on WSL. Now it should be fixed.

AmyrAhmady (approved):

Issue #1158, opened 2025-12-16
A couple of issues found with new NPCs


NexiusTailer:
**1. NPC_SetPos may remove bot from vehicle**
The first issue with `NPC_SetPos` native: when you teleported an NPC which was in car by FCNPC plugin, this bot always teleported right with his vehicle and continued to be in it on the new position set: https://github.com/ziggi/FCNPC/blob/master/src/CPlayerData.cpp#L997-L1004

But in omp implementation, an NPC may be either teleported without the vehicle or the vehicle may stay at the old position until re-stream, and it cannot be solved by using `SetVehiclePos` because this function doesn't work when NPC is a driver of the vehicle. So `NPC_SetPos` must always teleport an NPC who's a driver of this vehicle together with this vehicle, just like in FCNPC.

Test script where you can compare the behavior in-game: [FCNPC](https://pastebin.com/T67cefg5) / [omp NPCs](https://pastebin.com/CqTN2XcW) variant

**2. NPC doesn't call vehicle events like death or respawn**
If you apply `NPC_SetVehicleHealth` for a driver NPC and set any low health value for his vehicle like 150.0 or 0.0 when it starts firing and finally explode, still no `OnVehicleDeath` will be called, nor `OnVehicleSpawn` after a while too. Looks like this vehicle will explode and disappear only clientside, but not serverside. NPC will just stay at the same place without a car, moreover without his own death and respawn.

Test script where you can compare the behavior in-game: [FCNPC](https://pastebin.com/JanpXHkj) / [omp NPCs](https://pastebin.com/aY445Ni0) variant

**3. NPC_EnableInfiniteAmmo has no effect if NPC_SetAmmo wasn't used**
If an NPC has weapon with 0 ammo and you try to use `NPC_EnableInfiniteAmmo`, he still won't shoot any bullets:
```pawn
NPC_SetWeapon(npcid, random(3) + 22);
NPC_EnableInfiniteAmmo(npcid, true);
NPC_AimAtPlayer(npcid, playerid, .shoot = true);
```
but when you set this NPC at least 1 or more bullets with `NPC_SetAmmo`, he immediately start shooting:
```pawn
NPC_SetWeapon(npcid, random(3) + 22);
NPC_SetAmmo(npcid, 1); // or 5, or 999
NPC_EnableInfiniteAmmo(npcid, true);
NPC_AimAtPlayer(npcid, playerid, .shoot = true);
```
while FCNPC correctly handle even the first case

[...]

PR #1165, opened 2025-12-21
Fix issues with NPCs

SimoSbara:
This PR refers to #1158.

Status:
- [x] NPC_SetPos may remove bot from vehicle (solved in [6cac7da](https://github.com/openmultiplayer/open.mp/commit/6cac7da7847bd38a364e3e0079482c1c1c90029c))
- [x] NPC doesn't call vehicle events like death or respawn (it deserves another PR)
- [x] NPC_EnableInfiniteAmmo has no effect if NPC_SetAmmo wasn't used (done)

### DIFF
--- a/Server/Components/NPCs/NPC/npc.cpp
+++ b/Server/Components/NPCs/NPC/npc.cpp
@@ -631,6 +631,11 @@ uint8_t NPC::getWeapon() const
 
 void NPC::setAmmo(int ammo)
 {
+	if (ammo < 0)
+	{
+		ammo = 0;
+	}
+
 	ammo_ = ammo;
 
 	if (ammo_ < ammoInClip_)
@@ -1346,7 +1351,7 @@ int NPC::getWeaponActualClipSize(uint8_t weapon)
 			size *= 2;
 		}
 
-		if (ammo_ < size && !infiniteAmmo_)
+		if (ammo_ < size)
 		{
 			size = ammo_;
 		}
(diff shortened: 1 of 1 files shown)


SimoSbara:
After some debugging of the vehicle death event I observed that:
If a vehicle's health were set at **value <= 250** and it has never been occupied by a player in his lifetime (**common if you use it with a NPC**), the `Vehicle::setDead()` will never be called, and consequently `OnVehicleDeath()` will also be missed! This happens because the clients won't send the **VehicleDeath**'s packet (id 136) at the vehicle explosion, because they never entered the vehicle in the first place!

There is another way that the client can inform the vehicle's death to the player: [vehicle.cpp:304](https://github.com/openmultiplayer/open.mp/blob/cfc15ccf344f7602ff3cb64726262cb6c86d5e89/Server/Components/Vehicles/vehicle.cpp#L304) where the client can sync vehicles while onfoot. However, after multiple testing, it seems that the condition `if (health <= 0.f)` is never encoutered because after the explosion (health = 0), the client doesn't send anymore **PlayerUnoccupiedSync**'s packet (id 209).

Side note: `unoccupiedSync.SeatID != 0` could be removed because it seems like `SeatID` is always 0 when the player is on foot. (maybe it is used in some particular cases?)

Can anyone confirm/reject these observations?

SimoSbara:
@AmyrAhmady, I think it's better that I transform this PR for the infinite ammo improvement for NPC only. The vehicle death it's a separate question (from the observation that I had), and for the "NPC_SetPos issue" I saw that you already commit a fix.

What do you think?

NexiusTailer:
Thanks for your work and attention to my issue. I think the first item in the list can also be completely solved by adding `NPC_SetVehiclePos` (instead of any changes into `NPC_SetPos`), so there will be no differences with player manipulation natives.

SimoSbara:
I edited the PR, now it addresses the infinite ammo behaviour issue. 

> I think the first item in the list can also be completely solved by adding NPC_SetVehiclePos

I think it's better to separate it with another PR.

AmyrAhmady:
Merging this, I'll handle the other two in the rest

AmyrAhmady (approved):

PR #1179, opened 2026-01-05
Update player state when it exits spectate mode

AbdulOmarCR:
Update player state when it exits spectate mode, it was bugging without the state update.

### DIFF
--- a/Server/Source/player_impl.hpp
+++ b/Server/Source/player_impl.hpp
@@ -631,6 +631,12 @@ struct Player final : public IPlayer, public PoolIDProvider, public NoCopy
 			toSpawn_ = true;
 			spectateData_.type = PlayerSpectateData::ESpectateType::None;
 			spectateData_.spectateID = INVALID_PLAYER_ID;
+
+			// When client exits spectate mode it will attempt to request a class or to spawn
+			// but the server still thinks the it is in it's previous state.
+			// This can lead to those requests being rejected.
+			// This state update will fix that.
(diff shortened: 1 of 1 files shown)


NexiusTailer:
This change even worse since it breaks all serverside anticheats when they check player state skipping (illegal state changes like from none to onfoot skipping spawn etc).

The real solution is to make a "flag of being left from spec", to be added [here](https://github.com/openmultiplayer/open.mp/blob/master/Server/Source/player_impl.hpp#L631-L633) (where previously was the state messing kludge). So, thus:
1. When the server turns off spectating for player, it sets our new flag as true. As long as the player call `PlayerSpawnRPCHandler` (OnPlayerSpawn event), our flag will be reset to false
2. When we check players for the valid states suitable to call OnPlayerRequestClass and OnPlayerRequestSpawn (i.e. [here](https://github.com/openmultiplayer/open.mp/blob/master/Server/Components/Classes/classes_main.cpp#L156-L160) and [here](https://github.com/openmultiplayer/open.mp/blob/master/Server/Source/player_pool.hpp#L64-L68)), our flag can also be added as an exclusion for these two validity checks: they won't block those events **either** if player not spawned when calling them, ****or**** if our `beingLeftFromSpec` flag is true

NexiusTailer:
I plan to fix it myself in the next couple of days (because it was me who removed the previous problematic code and still did not provide the solution above instead), so it could probably stay opened until a new PR with all needed changes implementing actually working solution will be created.

Issue #1169, opened 2025-12-24
NPC component crash when Vehicles component absent


NexiusTailer:
**Describe the bug**
First of all it should not be critical nor really important for the most people because the issue only occurs when you purposely disable some of the default server components (and it's what you should not do for no reason). But since omp server always gave the ability to disable the most of the components shipped with the application, I consider that NPC should probably take the same behavior and be able to launch with no crashes in such case.

So, NPC component crashes at launch if you disable exactly Vehicles component. Disabling anything like Objects, Actors, Pickups and even Recordings won't produce any such issue.

**Screenshots and/or logs**
[omp_minidump.dmp](https://github.com/user-attachments/files/24333200/omp_minidump.dmp)

**Release version**
Latest release, v1.5.8.3079, tested on Windows server build

AmyrAhmady:
Fixed for the next release

PR #1182, opened 2026-01-07
NPC changes and fixes

AmyrAhmady:
NPC set position and set rotation changes for vehicles
Fix missing component crashes
Fix NPC_GetSkin return type

### DIFF
--- a/Server/Components/NPCs/NPC/npc.cpp
+++ b/Server/Components/NPCs/NPC/npc.cpp
@@ -192,13 +192,14 @@ Vector3 NPC::getPosition() const
 
 void NPC::setPosition(const Vector3& pos, bool immediateUpdate)
 {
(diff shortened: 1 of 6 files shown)


NexiusTailer:
Can confirm `NPC_GetSkin` is correct now 🐱‍🏍

NexiusTailer:
Can also confirm missing component issue (#1169) is solved

NexiusTailer:
I recently noticed one thing and as I don't wanna forget later, just leaving it here in the related PR:

We now have two new functions `NPC_SetVehiclePos` and `NPC_SetVehicleRot`, it's all setters. Worth thinking about also adding their counterparts, `NPC_GetVehiclePos` and `NPC_GetVehicleRot`.

PR #1187, opened 2026-01-12
Fix spawn when leaving spectator state

ksenonadv:
.. without breaking internal player state.
&
replaced deprecated OpenSSL SHA256 functions.


NexiusTailer:
Can now confirm it works correctly as in SA-MP server after some tests 👍
Thank you again!

Hual (approved):

PR #1188, opened 2026-01-15
Object editing types

xunder-matth:


NexiusTailer:
I also wonder if there's any validation now for calling object editing event (checking if player is in object edit mode). Can't check it rn in the sources as I from phone, but just leave this note here not to forget to test it later.

xunder-matth:
> I also wonder if there's any validation now for calling object editing event (checking if player is in object edit mode). Can't check it rn in the sources as I from phone, but just leave this note here not to forget to test it later.

Great call! There were checks for this, but they were shared between two modes. I was able to spoof the `EditObject` RPC while being in attach editing mode and vice versa. I have applied the fix for this.

Issue #1189, opened 2026-01-17
Server crashing when run gmx with (crashdetect plugin loaded, sscanf included and NPCs created)


denyktw:
**Describe the bug**
I'm hitting a crash after running gmx, It only happens when I have crashdetect loaded, sscanf included, and NPCs created.
If I remove any one of these three, the gmx works fine

**To Reproduce**
1. Download and compile this [bare.pwn](https://pastebin.com/vvSLUSq8) script
2. Add it into server config as gamemode, and add crashdetect, sscanf plugins into server config as well
3. Launch the server, and run gmx in server console
4. The server crashes after gmx

**Screenshots and/or logs**
Logs: [log.txt](https://github.com/user-attachments/files/24689789/log.txt)
.dmp: [omp_minidump.dmp](https://github.com/user-attachments/files/24689790/omp_minidump.dmp)

**Release version**
Latest release build 1.5.8.3079

And I also tested with the latest nightly build, and the issue still persists

denyktw:
Closing as it was fixed in b32634c

Issue #1167, opened 2025-12-23
Server crashing when rcon command gmx is used before new NPC is disconnected


ddanyk:
**Describe the bug**
As the title says, when I reload gamemode with gmx rcon command and there is one of the new NPCs connected, the server crash. This didnt happen in previous releases when I used old samp-npc.so.

**To Reproduce**
restart gamemode when NPC is connected

**Expected behavior**
no crash

**Screenshots and/or logs**
https://www.youtube.com/watch?v=V0z7fzZ8mDI

**Release version**
v1.5.8.3079

NexiusTailer:
#1160 probably related

AmyrAhmady:
I need you guys to test with latest nightly

NexiusTailer:
Likely should be solved in #1190

PR #1190, opened 2026-01-18
Call components reset after finishing unloading the main script.

ksenonadv:
Reset components after finishing unloading the main script.


NexiusTailer:
Surprisingly, now it's fixed (both #1167 and #1189) after the commit b32634c 🐱‍🏍

Hual (approved):

PR #1208, opened 2026-03-29
Update player_pool.hpp

Sranke:
Add a new feature of regex in nickname, so that your server can support player's nickname with different languages and some symbols!!

### DIFF
--- a/Server/Source/player_pool.hpp
+++ b/Server/Source/player_pool.hpp
@@ -1963,20 +1963,89 @@ struct PlayerPool final : public IPlayerPool, public NetworkEventHandler, public
 		{
 			return false;
 		}
+/*This new feature can supports ANSI() and supports native function"AllowNickNameCharacter"*/
 		try
 		{
 			std::string name_temp(name);
-			std::smatch result_of_regex;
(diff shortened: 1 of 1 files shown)


NexiusTailer:
There's already a native which can allow the specified characters in nicknames: [AllowNickNameCharacter](https://github.com/openmultiplayer/open.mp/blob/23b9bf6213afecb6be15098da05b291ca9250801/Server/Source/player_pool.hpp#L1973)

Sranke:
> There's already a native which can allow the specified characters in nicknames: [AllowNickNameCharacter](https://github.com/openmultiplayer/open.mp/blob/23b9bf6213afecb6be15098da05b291ca9250801/Server/Source/player_pool.hpp#L1973)

Hi! the [AllowNickNameCharacter] supports some symbol like"[]*$=@",which can display by ASCII encoding! But it doesn't support some symbols like "￡" which belong to ANSI (ANSI includes also character sets for multiple languages)!

NexiusTailer:
I'm about that you can still improve what's already exist instead of creating parallel mechanisms for the same functionality.

Sranke:
> I'm about that you can still improve what's already exist instead of creating parallel mechanisms for the same functionality.

OK!

PR #1213, opened 2026-04-19
vehicle paintjob reset on respawn, SeedCookie timing, and NPC vehicle death events

### LINKED ISSUE #1196: NPC doesn't call vehicle events like death or respawn
If you apply `NPC_SetVehicleHealth` for a driver NPC and set any low health value for his vehicle like 150.0 or 0.0 when it starts firing and finally explode, still no `OnVehicleDeath` will be called, nor `OnVehicleSpawn` after a while too. Looks like this vehicle will explode and disappear only clientside, but not serverside. NPC will just stay at the same place without a car, moreover without his own death and respawn.

Test script where you can compare the behavior in-game: [FCNPC](https://pastebin.com/JanpXHkj) / [omp NPCs](https://pastebin.com/aY445Ni0) variant

### LINKED ISSUE #1211: Rare connection issue: server responds to query but ignores connection packets (VPN or restarting router fixes it)
**Describe the bug**

A rare connection issue where some players can see the server in the SA-MP client and the server responds to query packets, but when they try to join they receive: The server didn't respond. Retrying...

This affects a very small percentage of players (approximately 2 out of 100). The issue appears to happen after a server restart. Packet capture shows that the server responds to query packets from the player but does not respond at all to connection/handshake packets. Interestingly, if the player connects through a VPN, the connection works immediately. When I restart the server, the players who previously couldn’t connect are able to join, but then some other players start experiencing the same issue and can’t connect anymore.

**To Reproduce**
The issue is difficult to reproduce consistently, but the observed behavior is:
1. The server appears online in the SA-MP client and responds to query packets.
2. A player attempts to join the server normally.
3. The player tries to connect.
4. They recieve message: The server didn't respond. Retrying...

**Expected behavior**
//

**Screenshots and/or logs**
(packets with size 4 are connection packets, while the others are query packets from same player)

Image

**Release version**
v1.5.8.3079

### LINKED ISSUE #1212: ChangeVehiclePaintjob bug
**Describe the bug**
When I set a vehicle's paintjob using the ChangeVehiclePaintjob function, the paintjob remains after the vehicle respawns, while other tuning components and the vehicle color are not preserved.

**To Reproduce**
1. Change a vehicle's paintjob using ChangeVehiclePaintjob
2. Respawn the vehicle
3. The paintjob remains

**Expected behavior**
After the vehicle respawns, the paintjob should be reset (just like other tuning components and colors).

**Screenshots and/or logs**
//

**Release version**
v1.5.8.3079

berkleblebiciii:
### 1. Reset paintjob on vehicle respawn (closes #1212)

Vehicle paintjob was not being cleared in `_respawn()` alongside other visual properties (mods, colours). After respawn the paintjob persisted, causing a desync between the vehicle's visual state and its server-side data.

**Fix:** Added `paintJob = 0` to `Vehicle::_respawn()` in `vehicle.cpp`.

---

### 2. Move SeedCookie after rakNetServer.Start() (closes #1211)

`SeedCookie` was being set before the RakNet server started, which could cause connection failures in certain scenarios.

**Fix:** Moved `SeedCookie` call to after `rakNetServer.Start()`.

---

### 3. Fix OnVehicleDeath / OnVehicleSpawn not firing for NPC-driven vehicles (closes #1196)

NPCs don’t send exit/death RPCs, so vehicles stay permanently occupied. This blocks the `onTick` respawn logic and prevents `OnVehicleDeath/Spawn`, leaving vehicles in a zombie state.

**Fix:** In `VehiclesComponent::onTick()` (`vehicles_impl.hpp`), detect the case where a dead vehicle's only occupant is an NPC driver and bypass the occupation check. The existing `_respawn()` already clears the driver pointer, so no further changes are needed.

### DIFF
--- a/Server/Components/LegacyNetwork/legacy_network_impl.cpp
+++ b/Server/Components/LegacyNetwork/legacy_network_impl.cpp
@@ -828,8 +828,6 @@ void RakNetLegacyNetwork::start()
 {
 	SAMPRakNet::Init(core);
 	SAMPRakNet::SeedToken();
-	lastCookieSeed = Time::now();
-	SAMPRakNet::SeedCookie();
 
 	playerFromRakIndex.fill(nullptr);
 
@@ -915,6 +913,14 @@ void RakNetLegacyNetwork::start()
 	{
 		rakNetServer.ReserveSlots(npcComponent->count());
 	}
+
+	// SeedCookie must be called after rakNetServer.Start() because Start() internally calls
+	// RakPeer::Disconnect() which resets RakNet's internal state. Seeding the cookie before
+	// Start() means the seeded value is discarded, leaving the cookie in an invalid state
+	// until the next onTick() re-seed. Clients connecting in that window fail the cookie
+	// check and their packets are silently ignored (issue #1211).
+	lastCookieSeed = Time::now();
+	SAMPRakNet::SeedCookie();
 }
 
 void RakNetLegacyNetwork::onTick(Microseconds elapsed, TimePoint now)
--- a/Server/Components/Vehicles/vehicle.cpp
+++ b/Server/Components/Vehicles/vehicle.cpp
@@ -707,6 +707,7 @@ void Vehicle::_respawn()
 	lastOccupiedChange = TimePoint();
 	timeOfSpawn = Time::now();
 	mods.fill(0);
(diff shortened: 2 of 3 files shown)


Joey251:
@Aesir1337 Thanks a lot, I implemented your SeedCookie solution and so far I haven’t noticed any complaints about connection issues.

NexiusTailer (approved):

PR #1216, opened 2026-04-28
Fix camera target player-object via dedicated extension

### LINKED ISSUE #1070: getCameraTargetObject -> invalid cast?
I've been trying to get the camera target of a player and was looking for `IPlayer->getCameraTargetPlayerObject()` but that doesn't appear to exist. I did however find that getCameraTargetObject does some player object related checking, but it looks like if it finds the player object the player camera is targeting, it tries to cast the `IPlayerObject*` to an `IObject*`

As far as I'm aware this is not a valid cast.

Shouldn't a separate `getCameraTargetPlayerObject` function be added?

```c++
// SDK declarations
struct IObject : public IBaseObject
{
    // (...)
};

struct IPlayerObject : public IBaseObject
{
    // (...)
};
```

```c++
IPlayerObject* player_object = data->get(cameraTargetObject_);
if (player_object)
{
	return reinterpret_cast<IObject*>(player_object);
}
```
https://github.com/openmultiplayer/open.mp/blob/5c4e67b4b3223b792f9cb44df54a809af047883d/Server/Source/player.cpp#L134C1-L138C5

bssth:
Closes #1070
Depends on openmultiplayer/open.mp-sdk#68

Implemented as an `IExtension` rather than a new virtual on `IPlayer` so
the `IPlayer` and `IPlayerObjectData` vtables stay byte-for-byte
unchanged and pre-built components remain binary-compatible

### DIFF
--- a/Server/Components/CAPI/Impl/Objects/APIs.cpp
+++ b/Server/Components/CAPI/Impl/Objects/APIs.cpp
@@ -821,10 +821,10 @@ OMP_CAPI(Player_GetSurfingPlayerObject, objectPtr(objectPtr player))
 OMP_CAPI(Player_GetCameraTargetPlayerObject, objectPtr(objectPtr player))
 {
 	POOL_ENTITY_RET(players, IPlayer, player, player_, nullptr);
-	IObject* object = player_->getCameraTargetObject();
-	if (object)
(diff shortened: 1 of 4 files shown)


bssth:
The SDK pin here is a temporary integration branch in the fork that
combines openmultiplayer/open.mp-sdk#65 (already required by
master) with our companion #68. Once both land, I can rebase to plain SDK master

PR #1218, opened 2026-04-28
vehicle custom

YagaPia:


NexiusTailer:
1. `‎artconfig.txt` has been deprecated in omp server
2. [Isn't it better to just add a more simple native from YSF?](https://github.com/IS4Code/YSF/blob/master/sampsvr_files/pawno/include/YSF.inc#L309-L310) As there need to be the changes from clientside as well for it to work, all the custom content would be added directly there by the server owners who maintain their custom clients. No reasons for downloading it in 0.3.DL style when you already modified your samp client to work with custom vehicle models.

PR #1230, opened 2026-06-07
CAPI Fix [closed due to branch issues]

### LINKED ISSUE #1199: Gangzone IDs are never recycled after destroy, causing pool exhaustion after 1024 creates
**Describe the bug**
Gangzones IDs are never returned to the pool after GangZone_Destroy() is called. While the destroy function returns success (true), the internal ID is not marked as available for reuse. This causes the gangzone pool to become permanently exhausted after 1024 total creations, even if only a handful of gangzones remain active.

**Expected behavior**
  - GangZone_Destroy() should mark the gangzone ID as available for reuse
  - After destroying a gangzone, its ID should be returned to a free pool
  - The system should support unlimited create/destroy cycles as long as active count < 1024
  - Only the active gangzone count should matter, not the total number of creates

**Actual Behavior**
  - GangZone_Destroy() returns success but does not free the ID
  - IDs increment monotonically from 0 to 1023 and are never reused
  - After 1024 total creations, GangZone_Create() always fails with nullptr and id=0
  - The gangzone pool becomes permanently exhausted until server restart
  - Server becomes unable to create gangzones even with 0 active zones

markski1:
Fixes #1199

Currently attempts to release legacy id after real id is released, which destroys the gz and causes the legacy id to be -1. This patch first stores the legacy id to avoid this, which is how it seems to already be done for pawn (Gangzones/Natives.cpp line 44 to 58)

Disclaimer: I've been unable to test this because I cannot make conan and PyYAML play ball with my windows python env. I'll try again when I have time, but this seems straightforwards enough.

### DIFF
--- a/Server/Components/CAPI/Impl/GangZones/APIs.cpp
+++ b/Server/Components/CAPI/Impl/GangZones/APIs.cpp
@@ -46,8 +46,9 @@ OMP_CAPI(GangZone_Create, objectPtr(float minx, float miny, float maxx, float ma
 OMP_CAPI(GangZone_Destroy, bool(objectPtr gangzone))
 {
 	POOL_ENTITY_RET(gangzones, IGangZone, gangzone, gz, false);
+	int legacyID = gangzones->toLegacyID(gz->getID());
 	gangzones->release(gz->getID());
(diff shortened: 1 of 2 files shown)


markski1:
Oops. Seems something I did on my fork closed this.

It is now at #1231

PR #1206, opened 2026-03-12
misc NPC natives and FCNPC parity

Knogle:
This adds a batch of missing FCNPC NPC natives and callbacks on the server side.

Added callbacks:
```
- OnNPCUpdate
- OnNPCChangeHeightPos
- OnNPCStreamIn
- OnNPCStreamOut
- OnNPCVehicleEntryComplete
- OnNPCVehicleExitComplete
- OnNPCVehicleTakeDamage
```

Added functions:
```
- NPC_SetQuaternion
- NPC_GiveQuaternion
- NPC_GetQuaternion
- NPC_GivePos
- NPC_GiveFacingAngle
- NPC_GiveHealth
- NPC_GiveArmour
- NPC_GiveAmmo
- NPC_GiveAmmoInClip
- NPC_GiveVelocity
- NPC_GiveSurfingOffsets
- NPC_GetClosestEntityInBetween
- NPC_SetPlaybackPath
- NPC_GetPlaybackPath
- NPC_SetWeaponInfo
- NPC_GetWeaponInfo
- NPC_SetWeaponDefaultInfo
- NPC_GetWeaponDefaultInfo
- NPC_SetMoveMode
- NPC_GetMoveMode
- NPC_SetMinHeightPosCall
- NPC_GetMinHeightPosCall
- NPC_ShowInTabListForPlayer
- NPC_HideInTabListForPlayer
```

This also adds the matching SDK exposure and keeps move mode documented as compatibility surface only where a real MapAndreas / ColAndreas backed implementation does not exist yet. For this case we have to consider some way to natively have Z data available for the server.

I've done the testing using the NPC test gamemode from @itsneufox https://github.com/itsneufox/NPC-bare-test-omp-gamemode

```
[2026-03-12T14:17:47+0100] [Info] [join] Knogle has joined the server (0:172.20.63.250)
[2026-03-12T14:17:53+0100] [Info] [NPC] NPC 99 has been created
[2026-03-12T14:17:53+0100] [Info] [connection] incoming connection: 127.0.0.1:50797 id: 99
[2026-03-12T14:17:53+0100] [Info] [npc:join] AutoBot_0 has joined the server (99:127.0.0.1)
[2026-03-12T14:17:53+0100] [Info] [NPC] NPC 99 has spawned
[2026-03-12T14:17:53+0100] [Info] [NPC-AUTO][player:0] Starting automated NPC parity test sequence.
[...]

PR #1249, opened 2026-08-02
Fix moving NPC vehicle pitch not synced

Fanorisky:
rotation sent in driver sync previously only used yaw, so moving npc vehicles looked flat instead of tilting toward their direction of travel. 
now pitch is calculated from position vs target position while moving_, then combined with yaw before sync.

### DIFF
--- a/Server/Components/NPCs/NPC/npc.cpp
+++ b/Server/Components/NPCs/NPC/npc.cpp
@@ -2222,15 +2222,10 @@ void NPC::sendDriverSync()
 	{
 		return;
 	}
-
 	uint16_t upAndDown, leftAndRight, keys;
 	getKeys(upAndDown, leftAndRight, keys);
-
 	uint16_t vehicleID = vehicle_->getID();
-
-	// Check if immediate update is needed (basic comparison for now)
 	bool needsImmediateUpdate = driverSync_.LeftRight != leftAndRight || driverSync_.UpDown != upAndDown || driverSync_.Keys != keys || driverSync_.Position != position_ || driverSync_.Rotation.q != rotation_.q || driverSync_.PlayerHealthArmour.x != health_ || driverSync_.PlayerHealthArmour.y != armour_ || driverSync_.VehicleID != vehicleID || driverSync_.Velocity != velocity_ || driverSync_.Health != vehicleHealth_;
-
 	auto generateDriverSyncBitStream = [&](NetworkBitStream& bs)
 	{
 		driverSync_.VehicleID = vehicleID;
@@ -2243,16 +2238,14 @@ void NPC::sendDriverSync()
 		driverSync_.PlayerHealthArmour.y = armour_;
 		driverSync_.Velocity = velocity_;
 		driverSync_.Health = vehicleHealth_;
-
 		driverSync_.Siren = uint8_t(useVehicleSiren_);
 		driverSync_.LandingGear = vehicleGearState_;
-
 		int model = vehicle_->getModel();
-		if (model == 520) // hydra model id
+		if (model == 520)
 		{
 			driverSync_.HydraThrustAngle = hydraThrusterDirection_;
 		}
-		else if (model == 537 || model == 538 || model == 570 || model == 569 || model == 449) // train part models
+		else if (model == 537 || model == 538 || model == 570 || model == 569 || model == 449)
 		{
 			driverSync_.TrainSpeed = vehicleTrainSpeed_;
 		}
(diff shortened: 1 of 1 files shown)


eakwarp:
@AmyrAhmady maybe its time?)
I tested the current implementation in all possible variants, and it's either this or NPC_SetQuaternion (from https://github.com/openmultiplayer/open.mp/pull/1206 ) is needed

Server/Components/NPCs/NPC/npc.cpp
```diff
@@ -2222,15 +2222,10 @@ void NPC::sendDriverSync()
 	{
 		return;
 	}
-
 	uint16_t upAndDown, leftAndRight, keys;
 	getKeys(upAndDown, leftAndRight, keys);
-
 	uint16_t vehicleID = vehicle_->getID();
-
-	// Check if immediate update is needed (basic comparison for now)
```
AmyrAhmady, line 2231: Why is the comment here removed?

Server/Components/NPCs/NPC/npc.cpp
```diff
@@ -2243,34 +2238,52 @@ void NPC::sendDriverSync()
...
 		driverSync_.LandingGear = vehicleGearState_;
-
 		int model = vehicle_->getModel();
-		if (model == 520) // hydra model id
+		if (model == 520)
 		{
 			driverSync_.HydraThrustAngle = hydraThrusterDirection_;
 		}
-		else if (model == 537 || model == 538 || model == 570 || model == 569 || model == 449) // train part models
+		else if (model == 537 || model == 538 || model == 570 || model == 569 || model == 449)
```
AmyrAhmady: Same here

AmyrAhmady (changes requested):
Some comments and some empty line breaks are removed for some reason

Server/Components/NPCs/NPC/npc.cpp
Fanorisky: ahh sorry bout that, its just a bad habbit.
i'll clean it up and push the fix

Server/Components/NPCs/NPC/npc.cpp
Fanorisky: fixed

AmyrAhmady:
instead of doing things raw, can you use 
```
const glm::vec3 delta = targetPosition_ - position_;

const float pitch = -atan2(delta.z, glm::length(glm::vec2(delta)));
const float yaw = glm::yaw(rotation_.q);

rotation_.q = glm::angleAxis(pitch, glm::vec3(1.0f, 0.0f, 0.0f)) * glm::angleAxis(yaw, glm::vec3(0.0f, 0.0f, 1.0f));
```
and put it in `NPC::move` instead of resetting it every sync packet broadcasting

Fanorisky:
@AmyrAhmady thanks, that's much cleaner .
adopted it

Server/Components/NPCs/NPC/npc.cpp
```diff
@@ -499,9 +499,18 @@ bool NPC::move(Vector3 pos, NPCMoveType moveType, float moveSpeed, float stopRan
 	{
 		front = (pos - position) / distance;
 		auto rotation = getRotation().ToEuler();
+		rotation.x = 0.0f; // Discard the pitch a previous drive move may have baked in, it would skew the facing angle
```
AmyrAhmady: I know this doesn't really change much, but since this line is meant to be for vehicles/driving only, let's move it in the `if` code block down there you made for `moveType_ == NPCMoveType_Drive`.

But remember it should be rotation_.x when you do.

AmyrAhmady (changes requested):
Please make sure you test it again in game after you made new changes, this would be my final change request and if you confirm it works in your test cases, I'll accept it

Server/Components/NPCs/NPC/npc.cpp
Fanorisky: `GTAQuat` only has a `glm::quat q` member, so `rotation_.x` won't compile. also, `rotation_.q.x` is a quaternion component, not an euler angle.

but I think this line actually belongs in the non-drive path rather than the drive one. inside the drive block, the quaternion gets overwritten completely by the `angleAxis` composition, so a stale pitch there shouldn't really matter.

the case that actually needs it is when an NPC drives uphill and then gets told to walk. `move()` calls `removeFromVehicle()` itself, and `sendFootSync()` writes the full quaternion, so the ped keeps walking around tilted by 26.5° even on flat ground.

moving it into the `if` would also mean adding a matching `else`, which basically just means having the same line in both branches.

Fanorisky: moved done

eakwarp:
@AmyrAhmady @Fanorisky Guys, what about this PR?)

